Issue #4, opened 2017-01-06
User emails should be unique


rossjones:
As we are allowing users to log in with an email address (in preference to a username) we will need to ensure that email addresses are only used by one user.  We can enforce this in our user_create override.

We will need to decide what to do about any existing duplicate emails.

PR #5, opened 2017-01-06
Add a user_auth endpoint for Alpha.

rossjones:
During Alpha we wish to be able to authenticate users without needing a
local model, and so here is an API endpoint, callable only by a sysadmin
which will return all of the user's detail on correct submission of
email and password.

The resulting dictionary also contains a list of the organisations that
the user is in.

One stumbling block is that we currently allow emails to be re-used by
different users.  For now we are just using the first such match, but
over time (as per #4) we need to resolve this (if any actually exist).

### DIFF
--- a/ckanext/datagovuk/action/get.py
+++ b/ckanext/datagovuk/action/get.py
@@ -1,0 +1,65 @@
+import ckan.logic.schema
+from ckan.plugins.toolkit import (
+    abort,
+    check_access,
+    navl_validate,
+    ValidationError,
+    get_action,
+)
+import ckan.lib.dictization.model_save as model_save
+import ckan.lib.dictization.model_dictize as model_dictize
+
+from ckan.model import User
+
+
+log = __import__('logging').getLogger(__name__)
+
+
+# defined as they are in ckan/action/create.py to save further hacks to the
+# function copied from there
+_check_access = check_access
+_validate = navl_validate
+
+
+def user_auth(context, data_dict):
+    '''Authenticates a user
+
+    You must be a system administrator to authenticate users.
+
+    :param email: the email address of the user
+    :type email: string
+    :param password: the password of the user
+    :type password: string
+
+    :returns: the newly created user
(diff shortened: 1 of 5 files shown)


ckanext/datagovuk/action/get.py
```diff
@@ -0,0 +1,65 @@
...
+_check_access = check_access
+_validate = navl_validate
+
+
+def user_auth(context, data_dict):
+    '''Authenticates a user
+
+    You must be a system administrator to authenticate users.
+
+    :param email: the email address of the user
```
davidread: The old system accepted "Username or e-mail address" so whatever we do about that on the front-end, it probably makes sense to login with a username too here.

ckanext/datagovuk/action/get.py
```diff
@@ -0,0 +1,65 @@
...
+    if not (email and password):
+        raise ValidationError(['email and password are both required'])
+
+    users = User.by_email(email)
+    user = users[0] if users else None
+
+    if (user is None) or \
+            (not user.is_active()) or \
+            (not user.validate_password(password)):
+        raise ValidationError(['There was a problem authenticating this user'])
```
davidread: Probably a good idea to add a log comment here explaining the problem, to keep track of trying usernames, failed password attempts, and attempts to use deleted users.

rossjones: Have added a ticket as I already merged.

ckanext/datagovuk/action/get.py
rossjones: I can imagine some users might not remember which email address they logged in with.  I'm not sure usernames as such will be a thing in the future, I've not heard any mention of them so far.  Let's see what the 'Register an account' flow looks like.

davidread: Yeah it'd just be to aid people transitioning from the old system, to continue logging in as they have done before. So I didn't see it as relevant to people registering afresh on the new system.

PR #12, opened 2018-05-18
Restyle the templates for DGU

brucebolt:
https://trello.com/c/oTeF4tCy/218-re-style-templates-minor-changes-s-m


ckanext/datagovuk/plugin.py
```diff
@@ -53,6 +54,12 @@ def package_types(self):
         # registers itself as the default (above).
         return []
 
+    # IRoutes
+
+    def after_map(self, map):
+        map.connect('harvest_index', '/harvest', action='index')
```
rossjones: In theory, the ckanext-harvest extension already has connected the route with a name.  I think this is fine though, it's more explicit, and isn't as hard to go and find.

PR #13, opened 2018-05-22
Template changes for publisher list

rossjones:
Removes the logo for publishers, and just puts them one on a line.  Also
removes the same logo from the dataset detail page.

Before:
publishers_-_dgu

After:
publishers_-_dgu-2


andrewgarner:
@rossjones can you add a before and after screenshot?

PR #14, opened 2018-05-22
Restrict registration to users with specific email patterns

brucebolt:
Valid email suffixes (or any other regex) are now specified in ckan.ini.  This is enforced on user registration through a validator.
https://trello.com/c/A09e4sKI/233-restrict-user-registration

The error messages need re-wording, this is on a separate card: https://trello.com/c/IoCaxZmP/246-adjust-the-text-wording-on-new-ckan


ckanext/datagovuk/schema.py
```diff
@@ -33,4 +33,14 @@ def show_package_schema(schema_dict):
                 toolkit.get_validator('ignore_missing')]
         })
 
-    return schema_dict
\ No newline at end of file
+    return schema_dict
+
+def user_new_form_schema():
+    from ckan.logic.schema import user_new_form_schema as default_user_new_form_schema
```
rossjones: It's okay for now, but we don't really have to do the imports here, they might be better at the top.  We did this historically but I'm not sure it's a great idea.  

One for next refactor I think.

PR #18, opened 2018-05-25
Rename org paths

rossjones:
Currently all of the publisher routes look like /organization, and this is wrong on at least 2 different levels.  This PR removes all of the routes for /organization* and replaces them with /publisher* so that any requests as follows will resolve properly....

```
url_for('organization_read', id='cabinet-office')
--> /publisher/cabinet-office
```

It also introduces some tests, and a helpful script (./run-tests.sh) which will save having to remember the ultimately forgetful incantation to launch nosetests.

See https://trello.com/c/JzzuMatU/271-overwrite-organization-route-with-publisher-in-ckan

### DIFF
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -1,6 +1,7 @@
 import ckan.plugins as plugins
 import ckan.plugins.toolkit as toolkit
 from ckan.lib.plugins import DefaultTranslation
+from ckan.config.routing import SubMapper
 
 import ckanext.datagovuk.auth as auth
 import ckanext.datagovuk.schema as schema_defs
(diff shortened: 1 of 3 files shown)


ckanext/datagovuk/plugin.py
```diff
@@ -79,11 +80,60 @@ def before_map(self, route_map):
...
                           controller=user_controller,
                           action='register')
-
+        route_map.connect('harvest_index', '/harvest', action='index')
         return route_map
 
     def after_map(self, route_map):
-        route_map.connect('harvest_index', '/harvest', action='index')
+        # Deletes all the organization routes
+        delete_routes_by_path_startswtih(route_map, '/organization')
```
andrewgarner: `startswith`?

rossjones:
@brucebolt Looks okay after a rebase now.

PR #19, opened 2018-05-25
Fixes, and tests that the inventory harvester assigns to an org

rossjones:
Changes the inventory harvester to allow the base class to do the
organisation assignment on our behalf.  Also introduces a test to
make sure this doesn't regress.

https://trello.com/c/gyuyaLlb/234-fix-inventory-harvester-org-assignment-m


ckanext/datagovuk/tests/test_inventory.py
```diff
@@ -0,0 +1,62 @@
...
+from ckanext.harvest.model import (HarvestObject, HarvestGatherError,
+                                   HarvestObjectError, HarvestSource,
+                                   HarvestJob)
+
+class TestInventory:
+
+    def setup(self):
+        db_setup()
+
+        self.sysadmin = factories.Sysadmin()
```
rossjones: You can use theses factories in test to avoid having to do a lot of setup.

ckanext/datagovuk/tests/test_inventory.py
```diff
@@ -0,0 +1,62 @@
...
+        self.publisher = factories.Organization()
+        self.source = HarvestSource(
```
rossjones: This should really be created with harvest_source_create action, but there are problems identifying the inventory harvester as it assumes the BaseHarvester.

ckanext/datagovuk/tests/test_inventory.py
```diff
@@ -0,0 +1,62 @@
...
+        harvest_object_id = result[0]
+
+        # Make sure we can create a dataset by running the import stage
+        harvest_obj = HarvestObject.get(harvest_object_id)
+        harvester.import_stage(harvest_obj)
+        assert(harvest_obj.package_id is not None)
+
+        # Get the newly created package and make sure it is in the correct
+        # organisation
+        pkg = toolkit.get_action('package_show')(
```
rossjones: This is the correct way to get package details from the DB, get_action() will return a function, and that takes a context, and then some parameters and returns a dictionary.

PR #21, opened 2018-05-29
Implement primary theme

brucebolt:
- Implements 'theme-primary' in package scheme
- Adds validator to restrict entry to limited selection and force selection of a value
- Value added to create/edit form and also to dataset view

Keys and values match those in new Publish: https://github.com/alphagov/datagovuk_publish/blob/63a6dcc203482623e734b5d7a62254259b2daa5c/db/seeds.rb#L6-L19

https://trello.com/c/uajsHBZr/244-add-theme-selection-to-ckan-manual-publishing-m


rossjones:
Might need to rebase on master

brucebolt:
Travis now passes after the fix in #22.

PR #36, opened 2018-06-06
Remove domain from URL

maxgds:
https://trello.com/b/a30DSRUe/dgu-q1-doing

This PR removes the default domain from the URL preview when creating datasets. We cannot give a full URL path (eg https://data.gov.uk/.....) as those urls contain a data ID that isn't available to CKAN at creation, but Gurinder advises that URL is still an accurate enough label for the field.

screen shot 2018-06-06 at 14 33 11

It also modifies the preview url for creating a new publisher - since this would be accessed in ckan the default domain is appropriate, but the slug preview name is now "publisher" instead of "organization".

We cannot easily modify the harvester url preview but the default state is appropriate.

### DIFF
--- a/ckanext/datagovuk/templates/organization/snippets/organization_form.html
+++ b/ckanext/datagovuk/templates/organization/snippets/organization_form.html
@@ -1,0 +1,20 @@
+{% ckan_extends %}
+
+  {% block basic_fields %}
+    {% set attrs = {'data-module': 'slug-preview-target'} %}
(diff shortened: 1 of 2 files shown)


andrewgarner:
Should we change the label to "Path"?

maxgds:
> Should we change the label to "Path"?

Gurinder said URL is fine, trying to change one individual instance of a translated term could be... fun? So I'd say no unless you're really keen for it.

brucebolt:
It's possible to change `_('URL')` to `_('Path')` to just update one instance.  If there is no match in the translation file, it will just use the name (in this case 'Path').

andrewgarner:
If Gurinder is happy with URL then 👍

PR #37, opened 2018-06-07
Add the contact/FOI fields

brucebolt:
Add contact/FOI fields to the publisher schema and also to the package schema.  Also removes the CKAN default contact/maintainer fields.

https://trello.com/c/7mvApeUY/305-add-correct-contact-fields


brucebolt:
I've added the missing fields

PR #55, opened 2018-06-12
Add Schema/Vocabulary selection

emmabeynon:
For https://trello.com/c/WMobiURu/303-add-schema-vocabulary

This adds a "Schema/Vocabulary" field to the dataset page and enables multiple schemas to be selected.

Confusingly, in old CKAN this field appears on the datafile/resource page but actually applies to the dataset as a whole - you cannot select "Schema/Vocabulary" for an individual datafile.  This is why we have added this field to the dataset page:

screen shot 2018-06-13 at 16 51 55

There is an outstanding alignment issue for the "Schema/Vocabulary" label which I have [created a card](https://trello.com/c/w1FlAoGh/375-fix-schema-vocabulary-label-alignment) for and we should probably think about whether the dropdown list itself should be wider in order to show more text.

We also display the chosen "Schema/Vocabulary" options on the dataset additional info page:

screen shot 2018-06-13 at 16 51 07


andrewgarner:
The current behaviour allows you to select multiple schemas. Is that supported here?

emmabeynon:
@andrewgarner yes I've added the `multiple` attribute to the select tag so that multiple schemas can be selected.  I need Bruce to check that it saves correctly as I've got some problems saving things in my local CKAN for some reason.

brucebolt:
This is showing for me as an autocomplete box, rather than the multiple select.
screen shot 2018-06-14 at 09 37 38

Fix should just be a case of removing `data-module="autocomplete"`.

emmabeynon:
@brucebolt thanks, I've removed that now. Can you check on your machine that it's working as expected for you?

PR #57, opened 2018-06-14
Substitute the user's username with their email address

brucebolt:
Updated the user register and edit methods to allow users to login and register with their email address, instead of a username.  Behind the scenes, the concept of a username still exists but is always set to equal the email address

https://trello.com/c/ELWatokd/345-allow-users-to-login-and-reset-password-using-email-address-or-username


brucebolt:
I think the method I proposed on Slack yesterday would be a cleaner solution.  Where we replace the authentication and password reset methods to lookup by username or email address.

It would remove all traces of the user's email from public sources, e.g. the API.

andrewgarner:
Okay. Shall we close this one then?

brucebolt:
Yes, I think so.

PR #65, opened 2018-06-18
Change Summary label to Description

emmabeynon:
For https://trello.com/c/UUJr13EC/372-change-label-for-summary-field-to-description

User research has shown that publishers are confused about the "Summary"
field, thinking it should be short and then looking for another field for a
full description, which doesn't exist. To solve this simply we change "Summary"
to "Description".

**Before**
screen shot 2018-06-18 at 15 04 44

**After**

screen shot 2018-06-18 at 14 55 46


andrewgarner:
What happens if you trigger a validation error for that field? Does it show "Description" too?

emmabeynon:
@andrewgarner I'm struggling to find a something that will trigger a validation error for this field.  It's not a required field so will validate if empty, for example.

brucebolt:
There is no validation on this field (https://github.com/ckan/ckan/blob/ckan-2.7.4/ckan/logic/schema.py#L154)

PR #66, opened 2018-06-19
Allow users to login with their username or email

brucebolt:
This overrides the default CKAN authenticator to lookup the user by email, if their input is not a valid username.

Also allows passwords to be reset by email, instead of username.

Updates the relevant labels from 'Username' to 'Username or Email'.


andrewgarner:
Can you rebase this @
?

brucebolt:
Rebased against master

PR #81, opened 2018-06-21
Remove "member" role

maxgds:
https://trello.com/c/jpPUupRZ/376-remove-member-role-when-adding-publishers

- Removes "member" as an option for a user when added to an org
- Tidies the form layout

### DIFF
--- a/ckanext/datagovuk/templates/organization/member_new.html
+++ b/ckanext/datagovuk/templates/organization/member_new.html
@@ -1,0 +1,90 @@
+{% extends "organization/edit_base.html" %}
+
+{% import 'macros/form.html' as form %}
+
+{% set user = c.user_dict %}
+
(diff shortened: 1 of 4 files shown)


ckanext/datagovuk/templates/organization/member_new.html
```diff
@@ -0,0 +1,90 @@
...
+{{ super() }}
+<div class="module module-narrow module-shallow">
+  <h2 class="module-heading">
+    <i class="fa fa-lg fa-info-circle"></i>
+    {{ _('What are roles?') }}
+  </h2>
+  <div class="module-content">
+    {% trans %}
+      <p><strong>Admin:</strong> Can add/edit and delete datasets, as well as
+        manage organization members.</p>
```
andrewgarner: Should this (and below) say "publisher members"?

maxgds: Fixed

PR #110, opened 2018-10-24
[DO NOT MERGE] Fix inventory XML harvester

brucebolt:
The inventory XML harvester was not working.

This PR correctly adds the owner organisation ID, which previously failed validation so no attempt was made to create the packages in the database.  It also imports a missing method `_match_resources_with_existing_ones` from legacy `ckanext-spatial` which is required to correctly assign harvested datasets to their correct CKAN package.

Dependent on: https://github.com/ckan/ckanext-harvest/pull/338

Trello card: https://trello.com/c/MTf9rBKK/3-get-harvesting-working-on-integration


bin/install-dependencies.sh
```diff
@@ -1,7 +1,7 @@
 #! /bin/bash
 
 venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckan/2.7/requirement-setuptools.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-harvest/v1.1.0/pip-requirements.txt)
+venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-harvest/c21c1627992cbcf0043b7b6873ff899f6f66767f/pip-requirements.txt)
```
andreagrandi: Hopefully my version won't break anything (I've tested it locally)

elliotcm: Your version of what sorry?

andreagrandi: @elliotcm this https://github.com/alphagov/ckanext-harvest it's the upstream one (forked yesterday) with the SSL patches the team applied in the last weeks/months. I will try to have them accepted upstream too, but in the mean time, if you want the SSL problem fixed, you can use our fork.

elliotcm: @andreagrandi Ah gotcha.  Should be fine to rebase that branch off the upstream commit mentioned here (or cherry-pick @brucebolt's commit), it's only a very tiny change.

PR #115, opened 2018-10-31
Remove truncation of resource names

brucebolt:
Resource names were being truncated at 50 characters on both the list of resources and also the actual resource itself.  This meant that a publisher could not differentiate between resources that have the same first 50 characters.

This PR removes truncation in both places and removes the green 'More' button, which confused users during testing and simply duplicates the behaviour of clicking the resource name.

Before:
screen shot 2018-10-04 at 10 39 24

screen shot 2018-10-31 at 12 09 29

After:
screen shot 2018-10-31 at 12 04 24

screen shot 2018-10-31 at 12 04 17

Trello card: https://trello.com/c/PU1T4N10/54-data-file-names-are-truncated

### DIFF
--- a/ckanext/datagovuk/templates/package/resource_read.html
+++ b/ckanext/datagovuk/templates/package/resource_read.html
@@ -1,4 +1,8 @@
 {% ckan_extends %}
 
+{% block resource_read_title %}
+  <h1 class="page-heading">{{ h.resource_display_name(res) }}</h1>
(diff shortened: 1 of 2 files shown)


brucebolt:
CKAN uses a templating system called Jinja2.  These two HTML files contain 'blocks' that override the 'block' of the same name in CKAN's core templates.

PR #117, opened 2018-11-05
Install deps in the same file

elliotcm:
This makes it easier to keep versions of eggs and their requirements in sync.

We shouldn't *need* to do this, but it seems the developers of these eggs haven't properly specified their own dependencies.

https://trello.com/c/pxc8vszb/89-get-ckan-running-in-staging-and-production

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ node {
     stage('Installing Packages') {
       sh("rm -rf ./venv")
       sh("virtualenv --no-site-packages ./venv")
-      sh("./bin/install-dependencies.sh")
+      sh("./bin/install-dependencies.sh ./venv/bin/pip")
     }
 
     stage('Tests') {
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -1,10 +1,21 @@
 #! /bin/bash
 
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckan/2.7/requirement-setuptools.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-harvest/c21c1627992cbcf0043b7b6873ff899f6f66767f/pip-requirements.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-dcat/master/requirements.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-spatial/master/pip-requirements.txt)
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -16,7 +16,7 @@ node {
     stage('Installing Packages') {
       sh("rm -rf ./venv")
       sh("virtualenv --no-site-packages ./venv")
-      sh("./bin/install-dependencies.sh")
+      sh("./bin/install-dependencies.sh ./venv/bin/pip")
```
andreagrandi: Is this script only ran by Jenkins or are we running it in our VM too?

bin/install-dependencies.sh
```diff
@@ -1,10 +1,21 @@
...
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-dcat/master/requirements.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-spatial/master/pip-requirements.txt)
-venv/bin/pip install -U $(curl -s https://raw.githubusercontent.com/alphagov/ckanext-s3-resources/master/requirements.txt)
-venv/bin/pip install -Ue 'git+https://github.com/ckan/ckan.git@ckan-2.7.4#egg=ckan'
-venv/bin/pip install -r venv/src/ckan/requirements.txt
-venv/bin/pip install -r requirements.txt
+pip=${1-'/usr/bin/env pip'}
+
+$pip install -U $(curl -s https://raw.githubusercontent.com/ckan/ckanext-harvest/c21c1627992cbcf0043b7b6873ff899f6f66767f/pip-requirements.txt)
+$pip install -U 'git+https://github.com/ckan/ckanext-harvest.git@c21c1627992cbcf0043b7b6873ff899f6f66767f#egg=ckanext-harvest'
```
andreagrandi: aren't we using our own fork on `https://github.com/alphagov/ckanext-harvest.git` ?

Jenkinsfile
elliotcm: The `install-dependencies` script is run by Jenkins, on the VM, and when deploying to production.  The `Jenkinsfile` is only run by Jenkins.

bin/install-dependencies.sh
elliotcm: Not as of this change, no.  See https://github.com/alphagov/ckanext-datagovuk/pull/117/files#diff-b4ef698db8ca845e5845c4618278f29aL2

Jenkinsfile
andreagrandi: oh... I see why we may have a few problems with pip in the VM then. I remember using Ansible in the past and the way to install packages is not to run an external script but I reckon we may want to reuse the same script for the VM, for Jenkins etc...

PR #119, opened 2018-11-15
Update data migration from API to SQL

brucebolt:
We are no longer using the CKAN API to migrate data between CKAN instances.  Instead, a database dump will be used and modified to support the update data models and schema.

This PR removes all the old import code, with everything rewritten as SQL to run over an imported database dump.

Trello card: https://trello.com/c/e3qWIvmc/48-implement-new-method-for-importing-data-from-old-ckan-to-new-ckan


brucebolt:
I used `git rm`

andreagrandi:
Very strange that GitHub is showing the file in the UI just like we only deleted the content. I remember it was showing deleted files in a different way. No worries.

brucebolt:
This will be run on a copy of the database in the "new world".  It won't be serving any production services at this point.

PR #123, opened 2018-11-30
Update CKAN homepage layout and content

brucebolt:
Removes the default CKAN homepage content and design.  Replaces with content similar to that planned for the new Publish app.

Before:
screen shot 2018-11-29 at 11 21 33

After:
screen shot 2018-11-29 at 15 15 14

Trello card: https://trello.com/c/ZhPoqwZp/123-update-ckan-homepage

### DIFF
--- a/ckanext/datagovuk/public/datagovuk.css
+++ b/ckanext/datagovuk/public/datagovuk.css
@@ -5,6 +5,7 @@
 body {
   font-family: "HelveticaNeue", "Helvetica Neue", "Arial", "Helvetica", sans-serif;
   font-size: 16px;
+  background: #FFF;
 }
 
 .masthead .navigation .nav-pills li a,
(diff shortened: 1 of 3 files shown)


ckanext/datagovuk/public/datagovuk.css
```diff
@@ -78,3 +79,41 @@ a.btn .icon-briefcase {
 label:after {
   content: "";
 }
+
+@media (min-width: 768px) {
+  .hero {
+    background: #128400;
```
elliotcm: The hero block is a different colour on smaller screens?

ckanext/datagovuk/public/datagovuk.css
```diff
@@ -78,3 +79,41 @@ a.btn .icon-briefcase {
...
+  }
+}
+
+.ie8 .hero,
+.ie7 .hero {
+  background: #128400;
+}
+
+.home-content {
+  color: #FFFFFF;
```
elliotcm: It's usually less cognitive load to use `white` and `black` or `#FFF` and `#000` for `#FFFFFF` and `#000000` but I don't mind much.

ckanext/datagovuk/public/datagovuk.css
brucebolt: Complete oversight!  Previously the tile pattern only showed if the screen was over that width.  We want the colour to be the same all the time.

PR #136, opened 2019-01-25
Make category an optional extra when showing orgs

elliotcm:
When this is mandatory in the DB to form schema it causes a 500 when
dumping the object to JSON if the extra is missing.

Because we have some categories missing in the DB, this change
allows us to render the orgs with the missing category but
requires a category to be provided if saving the org.


elliotcm:
Tested on integration.

PR #137, opened 2019-01-28
Better schema key error

elliotcm:
https://trello.com/c/BATbGtvS/742-some-schema-validation-issue-blocking-publishing-of-data-files-something-wrong-with-the-package-or-resource-extras


elliotcm:
@brucebolt It looks much the same?

https://github.com/ckan/ckan/blob/ckan-2.7.4/ckan/logic/validators.py#L816-L821

brucebolt:
You're right, it's exactly the same bar the error message

PR #138, opened 2019-01-31
Set a gunicorn request timeout from puppet

elliotcm:
This allows us to ensure that gunicorn has the same timeout
as nginx.

https://trello.com/c/2rnVLIgT/726-fix-timeout-on-organogram-publishing-recommended-%F0%9F%8D%90-for-dgu-learning

Relies on https://github.com/alphagov/govuk-puppet/pull/8614


elliotcm:
Do not deploy before puppet

PR #142, opened 2019-02-21
Disallow indexing by search engine bots

thomasleese:
The way we use CKAN is purely as an administration interface to manage the datasets which are rendered on data.gov.uk using a separate 'Find' application.

Therefore we don't want any robots to be indexing any of the pages in CKAN as they will only be accessible to people who have logged in.

Also we hope this will reduce the 5xx rate on CKAN as bots are frequently requesting a page which is raising a 5xx error and we'd rather stop them accessing this page rather than fixing the issue in CKAN itself.

This is based off [the official template from CKAN](https://github.com/ckan/ckan/blob/master/ckan/templates/robots.txt) but with modifications. As I understand it, templates in plugins will overwrite built in templates.

[Trello Card](https://trello.com/c/PonNGWyO/797-stop-search-engines-from-indexing-ckan)

### DIFF
--- a/ckanext/datagovuk/templates/robots.txt
+++ b/ckanext/datagovuk/templates/robots.txt
@@ -1,0 +1,5 @@
+{% ckan_extends %}
+
+{% block all_user_agents -%}
+Disallow: /
+{%- endblock %}


elliotcm:
@thomasleese @brucebolt Do we want to fully replace or extend like this?
https://github.com/alphagov/ckanext-datagovuk/commit/e5626082dd092b28e79b8e13360df02ab2c4721b

thomasleese:
Good idea @elliotcm, I've updated this PR to work by extending the original template.

thomasleese:
@elliotcm They strip whitespace to the left and right of the block lines, meaning you get

```
User-Agent: *
Disallow: /
```

and not

```
User-Agent: *

Disallow: /

```

PR #144, opened 2019-02-21
Remove old code relating to drupal passwords

thomasleese:
It was never used.


thomasleese:
@brucebolt Good point, I've completely removed this file now as I don't believe we need it.

thomasleese:
No idea why the tests aren't passing here, and don't have time to work on it :( so I'll close this for now.

PR #166, opened 2019-05-31
Script to fix existing organogram filenames

deborahchua:
The script will find filenames in S3 that don't end in `senior.csv`,
'junior.csv' or include `-posts-`, and replace them with the
correct file structure.

It also replaces the corresponding `url` in the database. This ensures
the url shown for its dataset in Find and CKAN matches, and shows
the organogram visualisation.

Trello card: https://trello.com/c/5KrmZ1tb/1025-fix-bug-to-show-visualisation-on-junior-post-organograms


bin/python_scripts/fix_organograms_s3_filenames.py
```diff
@@ -0,0 +1,264 @@
...
+def reindex_solr(dataset_name, dry_run):
+    paster_command = 'paster --plugin=ckan search-index rebuild {} -c /etc/ckan/ckan.ini'.format(dataset_name)
+
+    logger.info('SOLR reindex - %s command: %s', 'Will run' if dry_run else 'Running', paster_command)
+
+    if not dry_run:
+        try:
+            subprocess.call(paster_command, shell=True)
+        except:
+            exception = sys.exc_info()[1]
```
thomasleese: You can avoid having to do this, if you use `except exception:` in the line above.

bin/python_scripts/fix_organograms_s3_filenames.py
```diff
@@ -0,0 +1,264 @@
...
+            logger.error('Subprocess Failed, exception occured: %s', str(exception))
```
thomasleese: You can [pass the exception to the logger command like `error(..., exc_info=exception)`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) and avoid having to do the interpolation yourself and it'll print a nice stacktrace.

bin/python_scripts/tests/test_fix-organograms-s3-filenames.py
```diff
@@ -0,0 +1,56 @@
+#
```
thomasleese: In general, python filenames shouldn't have dashes in, otherwise they can't be imported.

bin/python_scripts/fix_organograms_s3_filenames.py
```diff
@@ -0,0 +1,264 @@
...
+    _format = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+    c_handler.setFormatter(_format)
+    f_handler.setFormatter(_format)
+    logger.addHandler(c_handler)
+    logger.addHandler(f_handler)
+
+    logger.info('====================================================================')
+
+
+def get_path_parts(obj, s3_ls, i):
```
thomasleese: Not a major thing and certainly not a blocker for this PR, but for this function you might find it easier to use [pathlib](https://docs.python.org/3/library/pathlib.html) rather than manually parsing the paths.

For example, you can do things like this:

```python
path = Path(obj.key)
dataset_name = path.parts[0]
filename = path_parts[-1]
directory = Path(*path.parts[:-1])

...

new_path = directory / new_filename
```

bin/python_scripts/fix_organograms_s3_filenames.py
kentsanggds: Will try that out

bin/python_scripts/fix_organograms_s3_filenames.py
kentsanggds: ok

bin/python_scripts/tests/test_fix-organograms-s3-filenames.py
kentsanggds: Though tests won't really be imported but will update

bin/python_scripts/fix_organograms_s3_filenames.py
kentsanggds: will try it out

kentsanggds:
We discovered another thing to fix, will need to use a specific timestamp as the  csv uploads had different timestamps between junior and senior csvs, will probably go with senior file timestamps.

PR #167, opened 2019-06-03
Update resource_form.html to add help on organograms

kentsanggds:
- also remove format as not needed on the upload page for central gov organograms

### DIFF
--- a/ckanext/datagovuk/helpers.py
+++ b/ckanext/datagovuk/helpers.py
@@ -288,8 +288,8 @@ def publisher_category():
 
 def activate_upload(pkg_name):
     pkg_dict = Package.by_name(pkg_name).as_dict()
-    pkg_extras = pkg_dict.get('extras')
-    return 'schema-vocabulary' in pkg_extras and pkg_extras['schema-vocabulary'] in ('538b857a-64ba-490e-8440-0e32094a28a7', 'd3c0b23f-6979-45e4-88ed-d2ab59b005d0')
+    pkg_extras = pkg_dict.get('extras', {})
+    return pkg_extras.get('schema-vocabulary') in ('538b857a-64ba-490e-8440-0e32094a28a7', 'd3c0b23f-6979-45e4-88ed-d2ab59b005d0')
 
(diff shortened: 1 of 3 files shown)


ckanext/datagovuk/helpers.py
```diff
@@ -297,3 +297,8 @@ def roles():
 
 def google_analytics_tracking_id():
     return config.get('ckan.google_analytics_tracking_id')
+
+def is_central_gov_organogram(pkg_name):
+    pkg_dict = Package.by_name(pkg_name).as_dict()
+    pkg_extras = pkg_dict.get('extras')
+    return 'schema-vocabulary' in pkg_extras and pkg_extras['schema-vocabulary'] == 'd3c0b23f-6979-45e4-88ed-d2ab59b005d0'
```
thomasleese: We could simplify this to `pkg_extras.get('schema-vocabulary') == 'd3c0b23f-6979-45e4-88ed-d2ab59b005d0'`.

ckanext/datagovuk/helpers.py
```diff
@@ -297,3 +297,8 @@ def roles():
 
 def google_analytics_tracking_id():
     return config.get('ckan.google_analytics_tracking_id')
+
+def is_central_gov_organogram(pkg_name):
+    pkg_dict = Package.by_name(pkg_name).as_dict()
+    pkg_extras = pkg_dict.get('extras')
```
thomasleese: I think we should use `.get('extras', {})` here so we get an empty dictionary if the key doesn't exist, otherwise in the next line we would get a `TypeError`. Or perhaps if we expect the key to exist should be just do `pkg_dict['extras']`?

ckanext/datagovuk/helpers.py
kentsanggds: True, I just copied the pattern from other code

ckanext/datagovuk/helpers.py
kentsanggds: I'll also update the other bits of code

PR #169, opened 2019-06-04
Show multiple validation errors

AlanGabbianelli:
Currently, even if the uploaded `xls` file has more than one validation issue, only one error message is displayed, making the process painful for the user. This PR fixes this problem, allowing the user to see all error messages at once.

It also makes the system accept `xlsx` files, instead of only `xls` files.

Trello card: https://trello.com/c/i57HcWI7/983-check-that-organogram-validation-is-working-as-expected

### DIFF
--- a/ckanext/datagovuk/action/create.py
+++ b/ckanext/datagovuk/action/create.py
@@ -69,9 +69,9 @@ def resource_create(context, data_dict):
 
             if errors:
                 context['session'].rollback()
-                error_summary = OrderedDict()
-                for i in range(0, len(errors)):
-                    error_summary[str(i + 1)] = errors[i]
+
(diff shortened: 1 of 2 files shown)


ckanext/datagovuk/action/create.py
```diff
@@ -67,7 +69,11 @@ def resource_create(context, data_dict):
 
             if errors:
                 context['session'].rollback()
-                raise ValidationError(errors)
+                error_summary = OrderedDict()
```
thomasleese: This could be written as:

```python
error_items = [(str(i + 1), error) for i, error in enumerated(error)]
error_summary = OrderedDict(error_items)
```

which might be more succinct.

kentsanggds: sure

PR #173, opened 2019-06-21
Add error reporting to Sentry

thomasleese:
Rather than using a custom CKAN plugin I've decided that the simplest way to enable Sentry error reporting is to implement it ourselves in this plugin.

All the CKAN Sentry plugins I looked at were a few years out of date and didn't support the new [Sentry SDK](https://github.com/getsentry/sentry-python).

Although this means that we now have to maintain this code, it's fairly simple and most of the work is done by the Python module itself.

Depends on https://github.com/alphagov/govuk-puppet/pull/9311.

[Trello Card](https://trello.com/c/Xv7JmC8g/1084-add-sentry-to-ckan-stack)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,8 +7,11 @@ ckanext-datagovuk
    What does it do? What features does it have?
    Consider including some screenshots or embedding a video!
 
+--------
+Features
+--------
 
-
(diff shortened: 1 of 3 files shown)


kentsanggds:
How do we configure sentry? 
Could something be added to the README for the configuration?

PR #192, opened 2019-08-16
Add PyCSW to the Procfile

thomasleese:
To run the PyCSW server, we need to have it configured in the Procfile so that we can generate an Upstart service for it.

This runs the `pycsw.wsgi` file (provided by the `pycsw` module) using gunicorn.

[Trello Card](https://trello.com/c/RudwUlvN/1157-update-govuk-puppet-to-deploy-ckan-with-pycsw)


thomasleese:
@cbaines Would you be able to take another look, it's changed a bit since your review.

cbaines:
> @cbaines Would you be able to take another look, it's changed a bit since your review.

Sure, this still looks good to me :+1:

PR #197, opened 2019-08-29
Remove PII from package search & show

kentsanggds:
## What
Removes PII determined by keys defined in PII_LIST in `ckanext/datagovuk/pii_helpers.py`

`package_search` and `package_show` were overridden by copying the code over to this repo. This is in line with how this repo overrides core ckan functions and how other extensions are doing this.

## Reference

https://trello.com/c/GrO3ZtIR/1228-remove-personal-data-from-ckan-api-endpoints


kentsanggds:
I did try that, but as it's not in a class we can't call `super`, calling it directly also doesn't give expected results.

PR #201, opened 2019-09-03
Remove users block from organization_show

kentsanggds:
## What

Remove the users block from `organization_show` endpoint to prevent exposure of hashed emails and user names.

I've carried out an end to end test locally up to the Find app and it appears to be working.

## Reference 

https://trello.com/c/ukgvMRf4/1246-remove-users-block-from-ckan-api-organizationshow-endpoint


kentsanggds:
Noticed on integration that it was showing some PII when fields set so have added additional step to remove PII

PR #202, opened 2019-09-03
Only update fields that are present in the data

kentsanggds:
## What

The api request by pycsw requests specific fields, we should not introduce additional fields if they are not provided.

### DIFF
--- a/ckanext/datagovuk/pii_helpers.py
+++ b/ckanext/datagovuk/pii_helpers.py
@@ -15,9 +15,12 @@ def remove_pii_from_api_search_dataset(data):
         json_data = json.loads(data)
 
         for element in json_data['results']:
-            element['data_dict'] = remove_pii_block(element['data_dict'])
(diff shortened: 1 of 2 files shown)


ckanext/datagovuk/pii_helpers.py
```diff
@@ -15,9 +15,12 @@ def remove_pii_from_api_search_dataset(data):
         json_data = json.loads(data)
 
         for element in json_data['results']:
-            element['data_dict'] = remove_pii_block(element['data_dict'])
-            element['validated_data_dict'] = remove_pii_block(element['validated_data_dict'])
-            element['extras'] = remove_pii(element['extras'])
+            if element.get('data_dict'):
```
thomasleese:
```suggestion
            if 'data_dict' in element:
```

This might be more idiomatic.

kentsanggds: ok, will make the change

PR #203, opened 2019-09-06
Remove PII from /api/3/search/dataset

kentsanggds:
## What

As puppet allows api calls with and without the version number we need to remove PII from the versioned API call.


kentsanggds:
After further investigation, found out that this isn't necessary

PR #207, opened 2019-09-20
Update package_schema to ignore harvest element

kentsanggds:
## What

Users were unable to update harvested datasets and got this error on the ckan publisher - 

```
The form contains invalid entries:

junk: The input field [('harvest', 0, 'value'), ('harvest', 0, 'key'), ('harvest', 1, 'value'), ('harvest', 2, 'value'), ('harvest', 1, 'key'), ('harvest', 2, 'key')] was not expected.
```

This was eventually traced to `augment_data` function which would check if any additional fields were added to the package dict. 

I was expecting `schema['harvest'] = [ignore]` to inform ckan to ignore that field but that didn't work, a work around was found that does work - `schema['harvest'] = {'_': ignore}`

An issue was created in ckan to see if there is a better way of doing this - https://github.com/ckan/ckan/issues/4989 or for the ckan team to improve their docs or fix the code to get it to work as expected.

## Reference

https://trello.com/c/zYenqUpG/1263-error-when-editing-data-source-on-datagovuk

### DIFF
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -73,7 +73,7 @@ class DatagovukPlugin(plugins.SingletonPlugin, toolkit.DefaultDatasetForm, Defau
             })
 
         from ckan.logic.schema import default_extras_schema
-        from ckan.lib.navl.validators import not_empty
+        from ckan.lib.navl.validators import ignore, not_empty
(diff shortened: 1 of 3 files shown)


kentsanggds:
I think I'm the only one who is actively working on CKAN these days.

PR #229, opened 2019-12-17
Update spatial sha in install-dependencies

kentsanggds:
## What

Update spatial sha to use commits from ckan spatial upstream


kentsanggds:
Reverted changes on Spatial

PR #230, opened 2019-12-17
Revert "Update dependencies to use latest harvest commit"

kentsanggds:
Reverts alphagov/ckanext-datagovuk#228

## What 

There was an issue with how rebasing was done on harvest, so will need to revert to previous sha


issyl0:
This has conflicts.

PR #231, opened 2019-12-18
Update harvest sha to add upstream commits

kentsanggds:
## What

This update uses a commit for the harvest branch which will be merged upstream.


issyl0:
Does this PR supersede #230?

PR #233, opened 2019-12-20
Update dependencies for dcat sha

kentsanggds:


bin/install-dependencies.sh
```diff
@@ -6,7 +6,7 @@ ckan_harvest_fork='alphagov'
 ckan_harvest_sha='5507adc82585a273c51eaebdf7e1d3bbe7d0f9dc'
 
 ckan_dcat_fork='alphagov'
```
thomasleese: Do we even need the fork anymore?

It looks like we're completely up to date with upstream: https://github.com/ckan/ckanext-dcat/compare/master...alphagov:master

kentsanggds: thats true, shall I just switch to upstream?

PR #237, opened 2020-01-14
Add code into ckanext-datagovuk to handle s3 resources

benjamineskola:
https://trello.com/c/NKJn1DwZ/1646-5-add-code-into-ckanext-datagovuk-to-handle-s3-resources


ckanext/datagovuk/plugin.py
```diff
@@ -273,8 +277,39 @@ def make_middleware(self, app, config):
...
+            logger.error(
+                "Required S3 config options missing. Please check if required config options exist."
+            )
+            raise Exception("Required S3 config options missing")
+        else:
+            upload.upload_resource_to_s3(context, resource)
+
+    def before_create(self, context, resource):
+        """Runs before resource_create. Modifies resource destructively to put in the S3 URL"""
+        print('***** before create')
```
kentsanggds: print should be removed

ckanext/datagovuk/plugin.py
```diff
@@ -273,8 +277,39 @@ def make_middleware(self, app, config):
...
+        self.before_create_or_update(context, resource)
+
+    def before_update(self, context, _, resource):
+        print('***** before update')
```
kentsanggds: print should be removed

ckanext/datagovuk/tests/test_package.py
```diff
@@ -1,3 +1,4 @@
+from mock import patch
```
kentsanggds: patch import is not needed

ckanext/datagovuk/tests/test_s3_resources_plugin.py
```diff
@@ -0,0 +1,150 @@
...
+    pkg = None
+    rsc = None
+
+    def __init__(self, _type):
+        self._type = _type
+
+    def __call__(self, *arg):
+        if self._type == "package_show":
+            return self.pkg or {"name": "test"}
+        elif self._type == "resource_show":
```
kentsanggds: I don't think we need a response for `resource_show` as its not called in `upload.py`

ckanext/datagovuk/tests/test_s3_resources_plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        plugin.before_create_or_update({}, resource)
+        assert not mock_abort.called
+        assert config.get("ckan.datagovuk.s3_url_prefix") not in resource["url"]
+
+    @patch("ckanext.datagovuk.upload.toolkit.abort")
+    @patch("ckanext.datagovuk.upload.toolkit")
+    def test_upload_s3_exception(self, mock_toolkit, mock_abort):
+        mock_toolkit.get_action = MockAction
+
+        s3 = botocore.session.get_session().create_client("s3")
```
kentsanggds: Why not just mock out `put_object` and introduce a `side_effect` so that an exception is raised? This should make the test a bit clearer.

ckanext/datagovuk/tests/test_s3_resources_plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        stubber = Stubber(s3)
+        stubber.add_client_error("put_object", expected_params={"Bucket": ANY})
+        stubber.activate()
+
+        plugin = DatagovukPlugin()
+        resource = self.resource.copy()
+
+        with assert_raises(Exception):
```
kentsanggds: We should probably target possible `boto3` specific exceptions rather and make it a catch all exception.

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,149 @@
+'''
+upload.py
+
+Contains functions that upload the resources/zipfiles to S3.
```
kentsanggds: Comment can be updated to just organograms as I don't think  we upload any other resources or zipfiles to s3.

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,149 @@
...
+    try:
+        logger.info("Uploading resource %s to S3" % resource.get('name', ''))
+        bucket.Object(s3_filepath).delete()
+        obj = bucket.put_object(Key=s3_filepath,
+                                Body=body,
+                                ContentType=content_type)
+        obj.Acl().put(ACL='public-read')
+        logger.info("Successfully uploaded resource %s to S3" % resource.get('name', ''))
+
+    except Exception as exception:
```
kentsanggds: We should catch a more specific S3 error rather than a catch all exception

ckanext/datagovuk/tests/test_s3_resources_plugin.py
```diff
@@ -0,0 +1,150 @@
...
+        "timestamp": "2020-01-09T12-11-41Z",
+        "datafile-date": u"",
+        "upload": upload,
+        "name": "2020-01-09 Organogram (Senior)",
+    }
+
+    @patch.dict("pylons.config", {}, clear=True)
+    def test_s3_config_exception(self):
+        plugin = DatagovukPlugin()
+        with assert_raises_regexp(Exception, "Required S3 config options missing"):
```
kentsanggds: Do we need to use `assert_raises_regexp`, will `assert_raises` work?

benjamineskola: I used `assert_raises_regexp` because the code raises a plain `Exception` with a message and I wanted to ensure it was the specific exception being raised. The alternative would be to raise a more specific exception, maybe `NameError`?

ckanext/datagovuk/plugin.py
```diff
@@ -273,8 +277,37 @@ def make_middleware(self, app, config):
...
+    def before_create_or_update(self, context, resource):
+        """before_create_or_update - our own function. NOT a CKAN hook.
+        Contains shared code performed regardless of whether we are
+        creating or updating.
+        """
+
+        # If resource is an API, don't do anything special
+        if resource.get("format") == "API":
+            return
+        elif "upload" not in resource or resource["upload"] == "":
```
thomasleese:
```suggestion
        elif not resource.get("upload"):
```

Because `None` and `""` are falsey.

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,150 @@
...
+                   + '/'
+                   + 'resources'
+                   + '/'
+                   + resource.get('timestamp', timestamp.strftime("%Y-%m-%dT%H-%M-%SZ"))
+                   + '-'
+                   + slugify(resource.get('name'), to_lower=True)
+                   + extension)
+
+    # If file is currently being uploaded, the file is in resource['upload']
+    if isinstance(resource.get('upload', None), cgi.FieldStorage):
```
thomasleese:
```suggestion
    if isinstance(resource.get('upload'), cgi.FieldStorage):
```

ckanext/datagovuk/tests/test_s3_resources_plugin.py
```diff
@@ -0,0 +1,142 @@
...
+import pylons.config as config
+from botocore.exceptions import ClientError
+from ckanext.datagovuk.plugin import DatagovukPlugin
+from mock import Mock, patch
+from nose.tools import assert_raises
+
+
+class MockAction:
+    pkg = None
+    rsc = None
```
thomasleese: I assume these are here because CKAN tries to read them?

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,150 @@
...
+    aws_region_name = config.get('ckan.datagovuk.s3_aws_region_name')
+    if aws_region_name:
+        s3 = boto3.resource('s3',
+                            aws_access_key_id=aws_access_key_id,
+                            aws_secret_access_key=aws_secret_access_key,
+                            region_name=aws_region_name)
+    else:
+        s3 = boto3.resource('s3',
+                            aws_access_key_id=aws_access_key_id,
+                            aws_secret_access_key=aws_secret_access_key)
```
thomasleese:
```python
kwargs = {
  "aws_access_key_id": aws_access_key_id,
  "aws_secret_access_key": aws_secret_access_key,
}

if aws_region_name:
    kwargs["region_name"] = aws_region_name

s3 = boto3.resource('s3', **kwargs)
```

I think this reads a bit better and avoids potential mistakes with the duplication of the access keys.

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,150 @@
...
+    timestamp = datetime.datetime.utcnow()
+    pkg = toolkit.get_action('package_show')(context, {'id': resource['package_id']})
+    s3_filepath = (pkg.get('name')
+                   + '/'
+                   + 'resources'
+                   + '/'
+                   + resource.get('timestamp', timestamp.strftime("%Y-%m-%dT%H-%M-%SZ"))
+                   + '-'
+                   + slugify(resource.get('name'), to_lower=True)
+                   + extension)
```
thomasleese: I wonder if this reads a bit better:

```python
filename = resource.get('timestamp', timestamp.strftime("%Y-%m-%dT%H-%M-%SZ")) \
    + '-' + slugify(resource.get('name'), to_lower=True) \
    + extension

s3_filepath = "/".join([pkg.get('name'), 'resources', filename])
```

Probably more personal preference at this point, but I thought it was worth mentioning.

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,150 @@
...
+    except ClientError as exception:
+        # Log the error and reraise the exception
+        logger.error("Error uploading resource %s from package %s to S3" % (resource['name'], resource['package_id']))
+        logger.error(exception)
+        if resource.get('url_type') == 'upload':
+            body.close()
+        raise exception
+
+    if resource.get('url_type') == 'upload':
+        body.close()
```
thomasleese: You could [use a `finally` block](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) on the `try` here to avoid the duplication.

```python
finally:
    if resource.get('url_type') == 'upload':
        body.close()
```

ckanext/datagovuk/upload.py
```diff
@@ -0,0 +1,150 @@
...
+    '''config_exists - checks for the required s3 config options'''
+    access_key = config.get('ckan.datagovuk.s3_aws_access_key_id')
+    secret_key = config.get('ckan.datagovuk.s3_aws_secret_access_key')
+    bucket_name = config.get('ckan.datagovuk.s3_bucket_name')
+    url = config.get('ckan.datagovuk.s3_url_prefix')
+
+    return not (access_key is None or
+                secret_key is None or
+                bucket_name is None or
+                url is None)
```
thomasleese: This might read better:

```python
return all([access_key, secret_key, bucket_name, url])
```

ckanext/datagovuk/plugin.py
benjamineskola: I'd tried this before and the tests failed — turns out it's because a FieldStorage can also be falsey, which is the fault of my test setup.

ckanext/datagovuk/plugin.py
```diff
@@ -286,16 +286,14 @@ def before_create_or_update(self, context, resource):
...
-        elif "upload" in resource and resource["upload"] == "":
+        elif resource.get("upload") is None:
             return
         elif not upload.config_exists():
             logger = logging.getLogger(__name__)
             logger.error(
                 "Required S3 config options missing. Please check if required config options exist."
             )
-            raise Exception("Required S3 config options missing")
+            raise NameError("Required S3 config options missing")
```
kentsanggds: Would LookupError be a better exception? - https://docs.python.org/2/library/exceptions.html#exceptions.LookupError

ckanext/datagovuk/plugin.py
```diff
@@ -286,16 +286,14 @@ def before_create_or_update(self, context, resource):
         # If resource is an API, don't do anything special
         if resource.get("format") == "API":
             return
-        elif "upload" not in resource:
-            return
-        elif "upload" in resource and resource["upload"] == "":
+        elif resource.get("upload") is None:
```
kentsanggds: Might be clearer like this - 

`elif 'upload' not in resource:`

benjamineskola: See https://github.com/alphagov/ckanext-datagovuk/pull/237#discussion_r366789215

PR #250, opened 2020-01-31
Patch guess_group_type for /publisher routes

deborahchua:
This was removed 2 years ago from core CKAN but is required for us
to route to organisations correctly. See ckan/ckan@e397e4c#diff-d48c332d8b78785e88afaf8406c9fff3

This patch also means we can defer the work on upgrading the pages
to flask views.

Trello card: https://trello.com/c/dljuAYRk/1691-5-fix-publisher-form-in-datagovuk-extension-in-docker-ckan-28-%F0%9F%8D%90


kentsanggds:
It's not in our code, basically the way that CKAN identifies the group type is via the URL, as it doesn't have publishers as a recognised group type it would render group templates rather than organization templates. There is a mention of this in the comment where the monkey patch happens. 

The commit which takes out the `guess_group_type` was taken out for CKAN 2.8, not for 2.7, which is why the current CKAN 2.7 is working.

PR #256, opened 2020-02-07
Fix upload timestamp

kentsanggds:
## What

The timestamps in the filenames were different for senior and junior csv files which meant that the frontend was not able to find the files. 

This fixes the timestamp to be the same for junior and senior csv files.

## Reference 

https://trello.com/c/Q7t5BErs/1742-fix-timestamp-when-uploading-organogram-in-ckanext-datagovuk

### DIFF
--- a/ckanext/datagovuk/action/create.py
+++ b/ckanext/datagovuk/action/create.py
@@ -15,6 +15,7 @@ import ckan.lib.dictization.model_dictize as model_dictize
 from ckan.logic.action.create import resource_create as resource_create_core
 from ckan.logic import get_or_bust
 from ckanext.datagovuk.lib.organogram_xls_splitter import create_organogram_csvs
+from ckanext.datagovuk.upload import upload_resource_to_s3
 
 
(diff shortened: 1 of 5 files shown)


cbaines:
> The timestamps in the filenames were different for senior and junior csv files which meant that the frontend was not able to find the files.
> 
> This fixes the timestamp to be the same for junior and senior csv files.

If this is useful information, it would be good to have it in the relevant commit messages.

ckanext/datagovuk/action/create.py
```diff
@@ -80,7 +81,10 @@ def resource_create(context, data_dict):
                 timestamp_str = timestamp.strftime("%Y-%m-%dT%H-%M-%SZ")
```
cbaines: I notice the commit message talks about timestamps, and this variable has timestamp in the name... is this the relevant timestamp?

kentsanggds: yes, thats the one that was in use by ckanext-s3-resources extension that we grabbed the code from

PR #262, opened 2020-02-13
Add script to replace .csv.csv with .csv

kentsanggds:
## What
This script reuses some code from fix_organograms_s3_filenames.py

It renames the s3 files and updates the urls with the correct path.

## Why
Fixing the path enables organogram visualisations to work.

## Reference 

https://trello.com/c/Q7t5BErs/1742-fix-timestamp-when-uploading-organogram-in-ckanext-datagovuk

### DIFF
--- a/bin/python_scripts/fix_organograms_s3_filenames_exts.py
+++ b/bin/python_scripts/fix_organograms_s3_filenames_exts.py
@@ -1,0 +1,195 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Requirements: 
+# in ./bin/python_scripts directory >
+#   sudo /usr/lib/venv/bin/pip install -r requirements.txt
+#
+# Set up environment from ckan.ini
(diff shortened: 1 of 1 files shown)


cbaines:
Is it known how files ended up with `.csv.csv` in the name?

kentsanggds:
Yes, it was accidentally introduced - https://github.com/alphagov/ckanext-datagovuk/pull/260/files#diff-9606a39e65a349d4e7319956bc47394dL86

That PR stops it from happening again has been deployed, this PR introduces a script to identify affected records and update the record and filename with the correct value.

cbaines:
> Yes, it was accidentally introduced - https://github.com/alphagov/ckanext-datagovuk/pull/260/files#diff-9606a39e65a349d4e7319956bc47394dL86
> 
> That PR stops it from happening again has been deployed, this PR introduces a script to identify affected records and update the record and filename with the correct value.

Ok, good to know :+1:

PR #267, opened 2020-02-18
Add password change email

kentsanggds:
## What

Send an email to the user whenever their password changes, either from the GUI, from a reset link or from the CLI.

## Why

There was an incident when the password was reset but no one knew how, this should add additional logs for when password get changed so that we have more visibility if someone attempts to gain unauthorised access.

## Reference

https://trello.com/c/EQcYxINw/1755-5-implement-email-alerts-to-users-when-their-password-is-changed-in-ckan

### DIFF
--- a/ckanext/datagovuk/lib/mailer.py
+++ b/ckanext/datagovuk/lib/mailer.py
@@ -1,0 +1,13 @@
+from ckan.lib import mailer
+import ckan.model as model
+
+log = __import__('logging').getLogger(__name__)
+
+
+def send_password_alert(user):
+    try:
+        body = "Your password has been changed, if you haven't done it yourself, let us know"
+        mailer.mail_user(user, 'Your CKAN password has changed', body)
+        log.info('Sent email to alert password change')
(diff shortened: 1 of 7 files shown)


ckanext/datagovuk/lib/mailer.py
```diff
@@ -0,0 +1,13 @@
+from ckan.lib import mailer
+import ckan.model as model
+
+log = __import__('logging').getLogger(__name__)
+
+
+def send_password_alert(user):
+    try:
+        body = "Your password has been changed, if you haven't done it yourself, let us know"
+        mailer.mail_user(user, 'Your CKAN password has changed', body)
```
nicholsj: Will users understand "CKAN"? Perhaps use "Your data.gov.uk publisher password..."?

kentsanggds: Password changes are on the CKAN publisher website, so I'm assuming that they know about it? Can easily change to `data.gov.uk publisher` though

PR #326, opened 2020-04-21
Only remove PII for json element with keys

kentsanggds:
## What

Sometimes the response is a list of strings and not json elements, so we should avoid trying to remove PII from the string otherwise we get a 500 server error.


risicle:
WFM

PR #344, opened 2020-05-06
Process api/3 endpoint to remove PII info

kentsanggds:
## What

We were handling the `/api/search/dataset` endpoint but not the `/api/3/search/dataset` endpoint , so this commit will also handle that.


risicle:
The issue wasn't 100% reproducible on a local instance, but this certainly had the effect of removing the fields where they _were_ being shown.

PR #375, opened 2020-06-03
Block POSTs on user/register

kentsanggds:
## What 

Although the registration form has been replaced, it was still possible to do a POST on the /user/register endpoint to create a user.

## Reference 

https://trello.com/c/7HEtFQiE/2010-3-investigate-and-fix-account-creation-disabled-but-not-really

### DIFF
--- a/ckanext/datagovuk/controllers/user.py
+++ b/ckanext/datagovuk/controllers/user.py
@@ -18,6 +18,7 @@ NotAuthorized = logic.NotAuthorized
 ValidationError = logic.ValidationError
 UsernamePasswordError = logic.UsernamePasswordError
 
+
 class UserController(UserController):
     def _new_form_to_db_schema(self):
         from ckanext.datagovuk.schema import user_new_form_schema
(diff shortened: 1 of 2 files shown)


ckanext/datagovuk/controllers/user.py
```diff
@@ -18,6 +19,9 @@
 ValidationError = logic.ValidationError
 UsernamePasswordError = logic.UsernamePasswordError
 
+ckan.controllers.user.UserController.original_new = ckan.controllers.user.UserController.new
+ckan.controllers.user.UserController.new = UserController.new
```
risicle, line 23: Why is the monkeypatching necessary? This is really confusing

risicle: The rest of our controller overrides simply re-declare the url mapping in `plugin.py`.

kentsanggds: Changed the way it's done now to do a super on the original function

PR #376, opened 2020-06-03
Fix button dropdown

maxgds:
https://trello.com/c/mgX7eAdl/168-harvest-source-delete-button-broken

The delete button on harvests should reveal two different delete options
when clicked, but is failing due to the implementation using a hash symbol
in the button link's href. This removes the hash and then adds styles to
give the dropdown options space on screen when it is clicked.

It seems the main issue is caused by some level of clash introduced between certain versions of bootstrap and jquery (see bug reports in the wild such as this: https://www.drupal.org/project/bootstrap_mint/issues/2957269)  and the bug reached beyond our own extension. Luckily the fix is extremely simple (remove the #) within a template we already override. 

Before the fix the delete button jumps the user to the top of the screen and does nothing, after the fix it reveals delete options thus:

Screenshot 2020-06-03 at 14 08 33

### DIFF
--- a/ckanext/datagovuk/templates/source/new_source_form.html
+++ b/ckanext/datagovuk/templates/source/new_source_form.html
@@ -90,13 +90,13 @@
     </div>
   {% endif %}
 
-  <div class="form-group form-actions">
+  <div class="form-group form-actions u-overflow-visible">
     {% block delete_button %}
       {% if data.get('id', None) and h.check_access('harvest_source_delete', {'id': data.id}) and not data.get('state', 'none') == 'deleted' %}
         {% set locale_delete = h.dump_json({'content': _('This will flag the source as deleted but keep all its datasets and previous jobs. Are you sure you want to delete this harvest source?')}) %}
         {% set locale_clear = h.dump_json({'content': _('Warning: Apart from deleting this source, this command will remove all its datasets, as well as all previous job reports. Are you sure you want to continue?')}) %}
   <div class="dropdown btn-group">
-    <a href="#" class="btn btn-danger dropdown-toggle" data-toggle="dropdown">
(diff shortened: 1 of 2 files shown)


ckanext/datagovuk/templates/source/new_source_form.html
```diff
@@ -90,13 +90,13 @@
...
 
-  <div class="form-group form-actions">
+  <div class="form-group form-actions u-overflow-visible">
     {% block delete_button %}
       {% if data.get('id', None) and h.check_access('harvest_source_delete', {'id': data.id}) and not data.get('state', 'none') == 'deleted' %}
         {% set locale_delete = h.dump_json({'content': _('This will flag the source as deleted but keep all its datasets and previous jobs. Are you sure you want to delete this harvest source?')}) %}
         {% set locale_clear = h.dump_json({'content': _('Warning: Apart from deleting this source, this command will remove all its datasets, as well as all previous job reports. Are you sure you want to continue?')}) %}
   <div class="dropdown btn-group">
-    <a href="#" class="btn btn-danger dropdown-toggle" data-toggle="dropdown">
+    <a href="" class="btn btn-danger dropdown-toggle" data-toggle="dropdown">
```
owenatgov, line 99: As this is no longer a link, I feel like it would make more sense for this to be a `button` element instead.

Besides that, looks and works a ok.

maxgds: Fixed.

owenatgov: Annoying nitpick: you missed the closing tag on line 102 😬

maxgds: Oops. Fixed.

PR #383, opened 2020-06-09
Add script to find invalid tags

kentsanggds:
## What

Add a script which finds invalid tags.

Shows the invalid tags, number of datasets using the tag and the validation error message.

This will be a one off script to help inform the next action to take with tags and the output file should be able to feed into the actual tag cleaning process.

## Reference 

https://trello.com/c/YJGBqhLv/2016-investigate-impact-of-restoring-defaults-schema-to-spatial-tags

### DIFF
--- a/bin/python_scripts/find_invalid_tags.py
+++ b/bin/python_scripts/find_invalid_tags.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Set up environment from ckan.ini
+# export POSTGRES_URL=<sqlalchemy.url from ckan.ini>
+#
+# Execute script like this -
+# python find_invalid_tags.py
+#
+
+import io
+import os
+import sys
+import psycopg2
+import subprocess
+
+import logging
+
+from ckan.logic.schema import default_tags_schema
+from ckan.lib.navl.dictization_functions import validate, Invalid
+
+POSTGRES_URL = os.environ['POSTGRES_URL']
+
+logger = logging.getLogger(__name__)
+connection = psycopg2.connect(POSTGRES_URL)
+
+
+def setup_logging():
+    _format = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+    logger.setLevel(logging.INFO)
+    c_handler = logging.StreamHandler()
+    c_handler.setFormatter(_format)
+    logger.addHandler(c_handler)
+
+    f_handler = logging.FileHandler('find_invalid_tags.log')
+    f_handler.setFormatter(_format)
+    logger.addHandler(f_handler)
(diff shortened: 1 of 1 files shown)


bin/python_scripts/find_invalid_tags.py
```diff
@@ -0,0 +1,109 @@
...
+import sys
+import psycopg2
+import subprocess
+
+import logging
+
+from ckan.logic.schema import default_tags_schema
+from ckan.lib.navl.dictization_functions import validate, Invalid
+
+POSTGRES_URL = os.environ.get('POSTGRES_URL')
```
thomasleese, line 22:
```suggestion
POSTGRES_URL = os.environ['POSTGRES_URL']
```

I'm guessing most of this script will fail if it's not set?

bin/python_scripts/find_invalid_tags.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    cursor.execute(sql)
+    return cursor
+
+
+def create_tag_list(rows):
+    with io.open("invalid_tags.txt", "w+", encoding="utf-8") as f:
+        for row in rows:
+            f.write(row)
```
thomasleese, line 58: I think `rows` is coming in as a string to this function, so it's writing each character one by one. `f.write(rows)` on its own should be the same, although I think it's probably better if we're passing a `list` into this function.

bin/python_scripts/find_invalid_tags.py
```diff
@@ -0,0 +1,109 @@
...
+    counter = 0
+    total_count = 0
+
+    for i, tag in enumerate(get_all_tags()):
+        logger.info('%d - %s', i, tag)
+        tag_name, count = tag
+        tag_name_utf8 = tag_name.decode('utf-8')
+        errors = validate_tag(tag_name_utf8)
+        if errors:
+            counter += 1
```
thomasleese, line 98: If `txt_rows` was stored as a `list` we could use `len(txt_rows)` in place of `counter` in the loop, similarly I think we could get rid of `total_count`.

bin/python_scripts/find_invalid_tags.py
kentsanggds: Yes it will fail, I guess better to fail early

bin/python_scripts/find_invalid_tags.py
kentsanggds: Thats true

bin/python_scripts/find_invalid_tags.py
kentsanggds: I can get rid of `counter` by using a list but need `total_count` as that accumulates the `count` returned by the validation function

PR #387, opened 2020-06-12
user dashboard template: update to work with ckan 2.8

risicle:
https://trello.com/c/e43yK9AF

Pretty straightforward.

### DIFF
--- a/ckanext/datagovuk/templates/user/dashboard.html
+++ b/ckanext/datagovuk/templates/user/dashboard.html
@@ -3,11 +3,11 @@
 {% block page_header %}
   <header class="module-content page-header hug">
     <div class="content_action">
-      {% link_for _('Edit settings'), controller='user', action='edit', id=user.name, class_='btn', icon='cog' %}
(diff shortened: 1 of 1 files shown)


owenatgov:
WFM ™️ 

One side note on this is that there's no way to get to the dashboard outside of that initial login nav or manually navigating to `/dashboard`. There's a "dashboard" link in the header but it goes to `/home`. The `header.html` template has this line that defines it:

```
<a href="{{ h.url_for(controller='user', action='dashboard_datasets') }}" title="{{ notifications_tooltip }}">
```
I don't know enough about ckan to interpret what's going on here.

If it's a hassle to sort this out as part of this PR then let me know and I'll stamp an approval onto this.

risicle:
Yeah I realize this - it's all a bit weird. You _should_ be directed to the `/dashboard/datasets` page following login or anything else that bounces via `/user/me`, but that's only because we seem to have made the decision to prefer `/dashboard/datasets` over `/dashboard` some time back. I just wanted it to Not Be Crashing™

owenatgov:
Gotcha. In that case I'll approve. I'll drop something in slack about this.

PR #388, opened 2020-06-12
Fix status field width

owenatgov:
## What
Adds bootstrap column classes to the label and input elements for the status form fields.

**Card:** https://trello.com/c/87ABxr5Q/198-add-data-state-bug

### Before
Screenshot_2020-06-11_at_16 02 29

### After
Screenshot 2020-06-12 at 15 34 10


owenatgov:
> WFM. Confusing why it's only shown in that context (not even in "edit dataset") 🤷

I think the intention is that This for before a dataset is published. I agree it's confusing and really we should review these journeys with more scrutiny down the road. In the mean time ta ta for the approve 👍

PR #415, opened 2020-07-09
Fix jenkins CI for master-2.8

risicle:
~This should eventually be the PR that fixes jenkins ci for the `master-2.8` branch. It is nowhere near ready yet, but I need to create a PR to be able to poke jenkins from.~

https://trello.com/c/zLLNaKtv

This allows our `master-2.8` branch to have CI tests pass. To do this I needed to change the setup of solr on some of the CI agents (`ci-agent-7` and `ci-agent-8`). This PR restricts this branch to running on those machines. It also changes the way the tests handle/expect resources to be set up on the agent's database and solr instance. Rather than expecting a single, fixed database/solr core (which caused various problems and potentially would have prevented simultaneous ci runs), it now creates a uniquely named postgres database and solr core before each run and removes them afterwards.

Though in the 2.8 branch we've moved to a "multicore" solr, with the ability to create & destroy cores at runtime, this still doesn't give us the same amount of flexibility as we have with postgres databases. We have to name a specific `configSet` to be used at core creation time (in this case `ckan28`) which the solr installation must know about (we put this in place with puppet currently). This `configSet` includes the schema that will be used for the core.

ckan 2.9 appears to be able to use another form of schema management for solr which is something a little closer to the flexibility of postgres databases but that's some way away.

I also realized we had been trying to fetch our custom ckan branches from the `ckan` fork rather than `alphagov` so fixed that.

Of course, it's still not perfect as CI because we're still pointing at moving targets in `install-dependencies.sh`, so two test runs could have different results depending on what gets merged to external repos.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ library("govuk")
 
 REPOSITORY = 'ckanext-datagovuk'
 
-node {
+node ('ci-agent-7 || ci-agent-8') {
 
   try {
     stage('Checkout') {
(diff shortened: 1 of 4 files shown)


risicle:
omg it's passing.

PR #420, opened 2020-07-16
Export invalid tag data to csv file

kentsanggds:
## What

Export the invalid tags with enough information to restore them if they are needed to be restored.

## Reference 

https://trello.com/c/oMXOxBxc/2031-export-invalid-tags-with-their-associated-package

### DIFF
--- a/bin/python_scripts/find_invalid_tags.py
+++ b/bin/python_scripts/find_invalid_tags.py
@@ -8,13 +8,7 @@
 # python find_invalid_tags.py
 #
 
-import io
-import os
-import sys
-import psycopg2
-import subprocess
-
-import logging
+import csv, io, logging, os, psycopg2, subprocess, sys
 
 from ckan.logic.schema import default_tags_schema
 from ckan.lib.navl.dictization_functions import validate, Invalid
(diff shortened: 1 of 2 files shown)


bin/python_scripts/find_invalid_tags.py
```diff
@@ -44,20 +44,50 @@ def get_all_tags():
...
+    rows = []
+    rows.append(u"index, tag, package_id, tag_id, vocabulary_id\n")
+    for i, tag in enumerate(get_tag_data()):
+        tag_name, package_id, tag_id, vocabulary_id = tag
+        if tag_name in invalid_tags:
+            rows.append(u"{}, {}, {}, {}, {}".format(
+                str(i + 1), tag_name, package_id, tag_id, vocabulary_id) + '\n')
+
+    with io.open("invalid_tags_data.csv", "w+", encoding="utf-8") as f:
+        f.writelines(rows)
```
thomasleese, line 86: It might be nicer to use the [`csv` module](https://docs.python.org/2/library/csv.html) here:

```python
with io.open("invalid_tags_data.csv", "w+", encoding="utf-8") as file:
    writer = csv.writer(file)
    writer.writerow(['index', 'tag', 'package_id', 'tag_id', 'vocabulary_id'])
    for i, tag in enumerate(get_tag_data()):
        ...
        writer.writerow([str(i + 1), tag_name, package_id, tag_id, vocabulary_id])
```

kentsanggds: Does it give us much benefit over just using the standard IO file writer?

thomasleese: I would argue it makes the code easier to read and nicer dealing with row structures rather than formatted strings. Probably one of the main benefits is that it'll properly support records that have got `,` characters in them.

kentsanggds: ok, thats a good enough benefit

PR #422, opened 2020-07-17
jenkins test setup: don't fail if ckan_test db can't be dropped

risicle:
This is probably just because it doesn't exist. that's fine.

This is a non-ideal way of handling resource setup anyway, and is replaced in our 2.8 branch. Fix this up temporarily in the 2.7 branch for now, because `master`'s CI is currently failing.


risicle:
(I'm just running the jenkins job a few times to make sure it didn't just get lucky)

risicle:
Hooray we've got this old `paster: command not found` error haunting us.

risicle:
The existing jenkins fail appears unrelated. Retrying.

PR #426, opened 2020-07-20
Add Dependabot config

ChrisBAshton:
[RFC 126] suggests configuring Dependabot to only
raise Pull Requests for the following three categories:

1. Security updates
2. Internal libraries
3. Framework libraries

Things have moved on slightly since the RFC; we're now using
[GitHub native] configs rather than the legacy `.dependabot`
approach. The syntax is slightly different to the example config
provided in the RFC but most of it is easily transferable.

There aren't many dependencies in this repo, so it doesn't
seem appropriate to define a strict 'allow' list for the
dependencies.

This commit gives us a blank canvas we can work with - we can
restrict dependencies further in a future commit if needed.

Trello: https://trello.com/c/uPoriyfJ/2049-add-dependabot-configuration-to-each-repo-blitz-pair

[GitHub native]: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates
[RFC 126]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-126-custom-configuration-for-dependabot.md

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: daily


ChrisBAshton:
Build has failed twice in a row, looks dependant on #422.

ChrisBAshton:
(Build was happy and passed!)

PR #435, opened 2020-07-24
install-dependencies.sh: specify --exists-action in pip calls

risicle:
https://trello.com/c/OxoHouSW

Otherwise pip is liable to prompt us for interactive input choosing an action, which is bad in a scripted scenario.

This choice should work reliably enough on remote installations yet shouldn't be too eager to overwrite a repo which a developer may have made local changes to (the `wipe` option risks doing this).

### DIFF
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -13,26 +13,28 @@ ckan_sha='ckan-2.8.3-dgu'
 
 pycsw_tag='2.4.0'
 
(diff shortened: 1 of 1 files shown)


risicle:
It seems this wasn't quite strong _enough_ for integration. It was strong enough for the CI agents, but, of course, the CI agents aren't the same as integration. I'm just going to cook up a stronger version of this...

PR #437, opened 2020-07-27
Upgrade CKAN 2.7.8 to add a number of fixes

kentsanggds:
## What

Upgrade CKAN to 2.7.8 for a number of fixes and also to fix the release to a particular ckan sha.


risicle:
I'm thinking how I'll test this - on the docker stack checkout the two repos to the specific commits by hand? Will try...

risicle:
Oh you merged it anyway I won't bother.

kentsanggds:
Think it should be ok, will test it on Integration and staging during the deployment process.

PR #439, opened 2020-07-28
Revert "Upgrade CKAN 2.7.8 to add a number of fixes"

kentsanggds:
Reverts alphagov/ckanext-datagovuk#437

## What 

Handling of PII was broken due to the way that extras were being output, this affects API calls from Publish app which will mean that Find will not be in sync with CKAN data until the revert or pii_helper code is updated to handle the change in data format.

No longer - 
```
            "extras": {
                "foi-email": "test-foi@example.com",
                "contact-name": "Test User",
                "contact-email": "test@example.com"
            },
```
but 
```
            "extras": [
                {
                    "key": u 'foi-email',
                    "value": "test-foi@example.com"
                },
                {
                    "key": u 'contact-name',
                    "value": "Test User"
                },
                {
                    "key": u 'contact-email',
                    "value": "test@example.com"
                }
            ],
```

### DIFF
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -9,8 +9,7 @@ ckan_dcat_sha='b757e5be643a17f08b1bb102348c370abee149d5'
 ckan_spatial_fork='alphagov'
 ckan_spatial_sha='46b706549aaf13a4ef2451d6185d7a90a75aeb0f'
 
-# CKAN release 2.7.8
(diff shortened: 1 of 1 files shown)


risicle:
I'm surprised they made this change in a point-release as it will have broken things for consumers of the REST API.

Still, this is something that would have been caught by my functional tests, and probably would only have exhibited itself if used in the "randomized testing" mode.

PR #443, opened 2020-07-29
Add remove_invalid_tags python script

kentsanggds:
## What

An update of the find_invalid_tags script to add information about the error message so that it can be used when running the remove_invalid_tags script to report to admin users that an invalid tag has been detected in a harvest source.

Also add the remove_invalid_tags script to remove the invalid tags based on the csv output from find_invalid_tags script

## Reference 

https://trello.com/c/pTkZmAPW/2008-3-remove-invalid-tags-from-ckan-database

### DIFF
--- a/bin/python_scripts/find_invalid_tags.py
+++ b/bin/python_scripts/find_invalid_tags.py
@@ -70,12 +70,14 @@ def create_tag_list(rows):
 def create_invalid_tag_data(invalid_tags):
     counter = 0
     with io.open("invalid_tags_data.csv", "wb") as f:
-        fieldnames = ['index', 'tag_name', "package_id", "tag_id", "vocabulary_id"]
+        fieldnames = ['index', 'tag_name', "package_id", "tag_id", "vocabulary_id", "errors"]
         writer = csv.DictWriter(f, fieldnames=fieldnames)
         writer.writeheader()
         for i, tag in enumerate(get_tag_data()):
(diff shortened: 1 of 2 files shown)


bin/python_scripts/find_invalid_tags.py
```diff
@@ -70,20 +70,23 @@ def create_tag_list(rows):
 def create_invalid_tag_data(invalid_tags):
```
thomasleese, line 70: I think it would be good to split the changes in this file into a separate commit.

bin/python_scripts/remove_invalid_tags.py
```diff
@@ -0,0 +1,152 @@
...
+    logger.addHandler(f_handler)
+
+    logger.info('====================================================================')
+
+
+def is_dev():
+    return '@db' in POSTGRES_URL
+
+
+def remove_invalid_tags(live=False):
```
thomasleese, line 42: It might be good to split this function into smaller functions, but I realise that this script will probably only be run once so perhaps it's not too important.

bin/python_scripts/find_invalid_tags.py
kentsanggds: ok, will do

bin/python_scripts/remove_invalid_tags.py
kentsanggds: I think I'll leave it for now

PR #446, opened 2020-07-31
Script to reindex solr using a list of package ids

kentsanggds:
## What

Target solr reindex to problem datasets.

### DIFF
--- a/bin/python_scripts/solr_reindex_package_ids.py
+++ b/bin/python_scripts/solr_reindex_package_ids.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Set the FILENAME_BASE to the txt file containing the list of package_ids
+#
+# Execute script like this -
+# python solr_reindex_package_ids.py
+#
+
+import os
+import sys
+import subprocess
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+FILENAME_BASE = "<filename for list of package_ids to reindex without the txt extension>"
+
+
+def setup_logging(log_to_file=False):
+    _format = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+    logger.setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


bin/python_scripts/solr_reindex_package_ids.py
```diff
@@ -0,0 +1,63 @@
...
+
+import os
+import sys
+import subprocess
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+FILENAME_BASE = "reindex_datasets-package_ids_28-31_07"
```
kevindew, line 18: Are you intending to commit this as a file? I notice line 37 seems to open "reindex_datasets-package_ids_28-31_07.txt"

bin/python_scripts/solr_reindex_package_ids.py
```diff
@@ -0,0 +1,63 @@
...
+def reindex_solr():
+    with open("%s.txt" % FILENAME_BASE, "r+") as f:
+        lines = f.readlines()
+        for i, line in enumerate(lines):
+            paster_command = 'paster --plugin=ckan search-index rebuild {} --config /var/ckan/ckan.ini'.format(line.strip())
+
+            logger.info('CKAN reindex %d/%d - Running command: %s', i+1, len(lines), paster_command)
+
+            try:
+                subprocess.call(paster_command, shell=True)
```
thomasleese, line 45: I think you'll want to use [`check_call`](https://docs.python.org/2/library/subprocess.html#subprocess.check_call) for it to raise an exception.

bin/python_scripts/solr_reindex_package_ids.py
kentsanggds: not really, I'll change it to some generic filename

bin/python_scripts/solr_reindex_package_ids.py
kentsanggds: ok, will change

PR #457, opened 2020-08-07
Setup test data

kentsanggds:
## What

To be able to run the CKAN functional tests we need to create test data, this can be done manually but it can take about 30 minutes to set it up and there's a chance that mistakes are made. Making it automated will make it easier to run the CKAN functional tests quickly.

## Reference 

https://trello.com/c/v7jpDYro/166-automate-creation-of-test-publisher-and-harvest-source

### DIFF
--- a/ckanext/datagovuk/lib/cli.py
+++ b/ckanext/datagovuk/lib/cli.py
@@ -1,8 +1,20 @@
 import csv
+import os
+import subprocess
 import sys
+import sqlalchemy
 
+from ckan import model
+from ckan.common import config
 from ckan.lib.cli import CkanCommand, query_yes_no
 from ckan.lib.mailer import create_reset_key
+
+from ckanext.harvest.harvesters.ckanharvester import CKANHarvester
+from ckanext.harvest.logic.action.create import harvest_job_create, harvest_source_create
+from ckanext.harvest.logic.schema import harvest_source_schema
+from ckanext.harvest.model import define_harvester_tables
+from ckanext.harvest.plugin import _create_harvest_source_object
+
 
 class PasswordResetsCommand(CkanCommand):
     '''Resets the passwords of all users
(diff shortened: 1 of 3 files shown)


ckanext/datagovuk/lib/cli.py
```diff
@@ -39,4 +51,238 @@ def command(self):
...
+    usage = '\n'.join(__doc__.split('\n')[1:])
+    max_args = None
+    min_args = None
+
+    def add_empty_field(dataset_id, fieldname, model):
+        field = model.PackageExtra(package_id=dataset_id, key=fieldname, value="")
+        model.Session.add(field)
+
+    def command(self):
+        if not [i for i in os.environ if i in ['CKAN_TEST_SYSADMIN_NAME', 'CKAN_TEST_SYSADMIN_PASSWORD', 'CKAN_INI']]:
```
risicle, line 76: I'm not sure this logic does what the error message suggests. I think you want:

```
if not all(os.environ.get(i) for i in ('CKAN_TEST_SYSADMIN_NAME', 'CKAN_TEST_SYSADMIN_PASSWORD', 'CKAN_INI')):
```

The `.get(i)` so that empty strings fall through and count as falsey.

ckanext/datagovuk/lib/cli.py
```diff
@@ -39,4 +51,238 @@ def command(self):
...
+        DELETE FROM member_revision WHERE group_id IN (SELECT id FROM "group" WHERE name = 'example-publisher-1');
+        DELETE FROM member WHERE group_id IN (SELECT id FROM "group" WHERE name = 'example-publisher-1');
+        DELETE FROM group_extra_revision WHERE group_id IN (SELECT id FROM "group" WHERE name = 'example-publisher-1');
+        DELETE FROM group_extra WHERE group_id IN (SELECT id FROM "group" WHERE name = 'example-publisher-1');
+        DELETE FROM group_revision WHERE name = 'example-publisher-1';
+        DELETE FROM "group" WHERE name = 'example-publisher-1';
+        DELETE FROM group_revision WHERE name = 'example-publisher-2';
+        DELETE FROM "group" WHERE name = 'example-publisher-2';
+        DELETE FROM "user" WHERE name = '%s';
+        ''' % os.environ.get('CKAN_TEST_SYSADMIN_NAME')
```
risicle, line 270: string-replacement in SQL strings :S

ckanext/datagovuk/lib/cli.py
kentsanggds: I've tested my code and it appears to work but I think your code is clearer so will implement that

risicle: Well from the looks of it what yours does is pass if _any_ of those three is set. Unless ... that's what you wanted? The error message is ambiguous.

ckanext/datagovuk/lib/cli.py
kentsanggds: ok, can do that

PR #472, opened 2020-08-19
Add alert banner to all pages

owenatgov:
## What
Adds an alert banner that appears on all pages of ckan with the following message:

> There will be a publishing freeze from 06:30am on 27 August to 11:59pm to 28 August. You won't be able to publish, edit or delete datasets while we upgrade the data publishing tool.
There will not be any downtime for visitors – only publishing will be affected.

This alert banner uses [the bootstrap alert warning component](https://getbootstrap.com/docs/4.0/components/alerts/).

What it looks like:
Screenshot 2020-08-19 at 14 26 34
Screenshot 2020-08-19 at 14 26 13

## Why
This is to alert users that ckan will be down next Thursday (27th August) whilst the DGU upgrade team move ckan from 2.7 to 2.8.

**Card:** https://trello.com/c/0VQ2mvRX/262-add-a-banner-for-advanced-warning-of-downtime

### DIFF
--- a/ckanext/datagovuk/public/datagovuk.css
+++ b/ckanext/datagovuk/public/datagovuk.css
@@ -99,6 +99,7 @@ p.home-text {
 
 .home-footer {
   color: #000;
+  background: #ffffff;
 }
 
 h1.home-footer-heading {
@@ -110,3 +111,6 @@ p.home-footer-text {
   line-height: 1.5;
 }
 
+.homepage {
+  background: #128400;
+}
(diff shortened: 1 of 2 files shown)


kentsanggds (approved):
LGTM, I'll check with Martin when to deploy and then will merge it through

kentsanggds:
Have noticed a minor thing with the homepage in that its missing some margin/padding, but it's only a small thing and still good to go

owenatgov:
> Have noticed a minor thing with the homepage in that its missing some margin/padding, but it's only a small thing and still good to go

Is this related to the lack of top margin on the homepage? I tried to get around this but it's very tricky with how it's set up. I might be able to hack something together if it's a problem?

kentsanggds:
Nope, its fine, only temporary until 2.8 gets released and Martin has ok'ed it

PR #477, opened 2020-08-24
Update alert banner to make it clearer

kentsanggds:
## What

Make it clearer that CKAN is unavailable but data.gov.uk is still working as normal.


owenatgov (approved):
⚠️ 🚀

PR #494, opened 2020-09-03
Ckan accessibility statement

owenatgov:
## What
Adds an /accessibility route which links to a plain content page containing the signed off ckan accessibility statement.

## Why
This is necessary to comply with impending public sector accessibility legislation. We can't fix all the accessibility issues we found before September 23rd, so we've written an accessibility statement committing to fix these issues.

**Card:** https://trello.com/c/ESi2dbOz/272-write-up-accessibility-statement-on-ckan


kentsanggds (approved):
👍

PR #504, opened 2020-09-09
Don't push data errors up to Sentry

kentsanggds:
## What

Re-enable sentry errors for Staging and Production but as data errors are already reported to publishers, pushing these errors up to Sentry will cause the rate limit to be hit preventing other apps from reporting errors. So prevent them from being sent to Sentry as they are already recorded else where and presented to the publishers.

## Reference 

https://trello.com/c/xLnwdSng/274-investigate-errors-in-sentry-and-prevent-unnecessary-ones-from-being-pushed-up

### DIFF
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -325,7 +325,7 @@ class DatagovukPlugin(plugins.SingletonPlugin, toolkit.DefaultDatasetForm, Defau
     ]
 
     def before_send(self, event, hint):
-        return None if [i for i in ['localhost', 'integration', 'staging'] if i in config.get('ckan.site_url')] or \
+        return None if [i for i in ['localhost', 'integration'] if i in config.get('ckan.site_url')] or \
             any(s in event['logentry']['message'] for s in self.IGNORED_DATA_ERRORS) \
             else event
 
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -315,11 +315,19 @@ class DatagovukPlugin(plugins.SingletonPlugin, toolkit.DefaultDatasetForm, Defau
 
     # IMiddleware
 
+    # Ignore these data errors on sentry as these would already be reported to publishers to resolve or raise
+    IGNORED_DATA_ERRORS = [
+        "Found more than one dataset with the same guid",   # DCat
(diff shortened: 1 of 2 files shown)


ckanext/datagovuk/plugin.py
```diff
@@ -315,11 +315,19 @@ def get_helpers(self):
...
+        "was aborted or timed out",                         # Harvest
+        "Too many consecutive retries for object"           # Harvest
+    ]
+
     def before_send(self, event, hint):
-        # disable sentry while CKAN is processing objects which have timed out due to upgrade
-        return None
-        # return None if [i for i in ['localhost', 'integration', 'staging'] if i in config.get('ckan.site_url')] \
-        #     else event
+        return None if [i for i in ['localhost', 'integration', 'staging'] if i in config.get('ckan.site_url')] or \
```
ChrisBAshton: Could you explain in the commit why we're allowing errors in Production here?

(Or better still, just add 'production' to this list for now, and then remove it from the list in a later commit),

ckanext/datagovuk/plugin.py
```diff
@@ -325,7 +325,7 @@ def get_helpers(self):
     ]
 
     def before_send(self, event, hint):
-        return None if [i for i in ['localhost', 'integration', 'staging'] if i in config.get('ckan.site_url')] or \
+        return None if [i for i in ['localhost', 'integration'] if i in config.get('ckan.site_url')] or \
```
ChrisBAshton, line 328: Same question here - why enable it in staging? I would assume you want to enable it in staging but _disable_ it in production, in order to try this out...?

ChrisBAshton (commented):
Code looks good 👍 just a couple of questions.

ckanext/datagovuk/plugin.py
kentsanggds: Thats the intention, we want sentry errors from Production in this commit

ckanext/datagovuk/plugin.py
kentsanggds: We also want errors from Staging, originally I disabled Staging errors first as they were more numerous due to data sync causing lots of data errors being thrown but have now mitigated that by ignoring data errors.

ChrisBAshton (approved):

PR #525, opened 2020-09-24
Restore public directory for organogram templates

benthorner:
https://govuk.zendesk.com/agent/tickets/4247650

Our Nginx config allows access to files in the "public" directory
of the deployed repo, but the expected files were located several
layers deeper. This moves the directory to the top-level, so it
becomes accessible again, and removes the other files that appear
to be unnecessary - only the organogram template has been missed.

https://github.com/alphagov/govuk-puppet/blob/7aba15f34d9ff92d59666786ff5c56eae5d0bcdb/modules/nginx/manifests/config/vhost/proxy.pp#L85


JonathanHallam (approved):
Looks good to me 👍 Good spot

PR #531, opened 2020-09-29
Ignore CSW identifier already used: Spatial plugin

kentsanggds:
## What

Ignore errors where CSW identifier is already used being sent to Sentry as users will already have been informed in their harvest logs and it is more likely a data issue rather than a system error.


benthorner (approved):
Looks good to me, although I'm not familiar with the errors. Thanks for reigning them in!

PR #545, opened 2020-10-16
Block package search calls if matching block_tags

kentsanggds:
## What

Block calls to package search if they match all block tags and log a warning when it does this.

### DIFF
--- a/ckanext/datagovuk/ckan_patches/__init__.py
+++ b/ckanext/datagovuk/ckan_patches/__init__.py
@@ -4,3 +4,4 @@
 from cli import *
 from logic import *
 from model import *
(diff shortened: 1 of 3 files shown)


risicle:
Mind is a little blown that either CKAN or solr doesn't handle this for us.

risicle (approved):
Definitely looks like it does what it says it does. Whether that is the right thing to do here, I don't know.

risicle:
Ah my reviews don't mean anything anymore.

kentsanggds:
Solr upgrade does handle this so this is extra insurance.

thomasleese (approved):

PR #583, opened 2020-11-30
Update styling of active navigation items

owenatgov:
## What
Amend the styling of navigation item highlighting. Specifically:

- Add a `/` to the end of the `href` selector on navigation links
- Remove some styles that are no longer necessary

## Why
During the upgrade from 2.8 to 2.9, the links between sections in the ckan UI changed to include a trailing slash on the links. This broke the styling we currently have for links that are specified via the `href`. This feature is useful to communicate to users where they are within the ckan service.

This level of specificity in our CSS is less than ideal, however we aren't in full control of this setup as the navigation is handled within ckan (not this extension).

I removed the additional calls to `li.active a` as I don't envision a situation where we would ever have 2 "active" `li`'s at any one time. We also don't have a site usage page anymore that's accessible from the header navigation so I've removed this as well.

## Additional context
Tests that theoretically aren't effected by this change are failing intermittently here, however this is expected as it's part of an ongoing upgrade on this app from python 2 to python 3 (hence the merging-into branch name). The tests working against python 3 and the test suit set up on python 3 are being handled in a separate stream of work by @kentsanggds.

### Before
Screenshot 2020-11-30 at 15 52 54

### After
Screenshot 2020-11-30 at 15 53 04

### DIFF
--- a/ckanext/datagovuk/public/datagovuk.css
+++ b/ckanext/datagovuk/public/datagovuk.css
@@ -18,14 +18,11 @@ body {
   background-color: #003647;
 }
 
-.dataset .masthead .navigation .nav-pills li a[href="/dataset"],
-.dataset .masthead .navigation .nav-pills li.active a[href="/dataset"],
-.publisher .masthead .navigation .nav-pills li a[href="/publisher"],
-.publisher .masthead .navigation .nav-pills li.active a[href="/publisher"],
-.harvest .masthead .navigation .nav-pills li a[href="/harvest"],
-.harvest .masthead .navigation .nav-pills li.active a[href="/harvest"],
-.site .masthead .navigation .nav-pills li a[href="/site-usage"],
-.site .masthead .navigation .nav-pills li.active a[href="/site-usage"] {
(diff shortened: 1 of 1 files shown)


ChrisBAshton (commented):
The code LGTM, but could you add all that lovely PR context to the commit message itself?

(Hopefully a rebase & force push will fix the build too, which seemed to fail intermittently).

owenatgov:
@ChrisBAshton Good shout. Updated the PR description with some more details. I'm going to make the case for not trying to get tests running here as they're currently unpredictable and `python3` is still a pre-release branch.

ChrisBAshton:
Fair enough RE the tests, but just to reiterate, would be good if you could edit the commit itself so that the context is captured independently of being on GitHub.com (it's currently only 1 line long and someone with the repo checked out locally would have to dig through GitHub to find out the 'why' behind the change).

owenatgov:
@ChrisBAshton Ahhhhh my bad! Completely misunderstood. Re-pushed the commit with the `why` from the PR.

ChrisBAshton (approved):

PR #623, opened 2021-01-19
Fix jenkins tests for CKAN 2.9 running on python 3

kentsanggds:
## What

Existing tests for ckanext-datagovuk are run on python 2.7, in order to get the tests to pass on Jenkins the venv needs to be created for python 3. 

Also nosetests have been removed as the tests have been migrated to pytests and they have been moved in order to show test coverage.

## Reference

https://trello.com/c/VrPqMpYM/2286-enable-the-ci-test-pipeline-to-run-against-ckan-29-on-python-3


bin/install-dependencies.sh
```diff
@@ -19,6 +19,10 @@ pipopt='--exists-action=b'
 
 $pip install $pipopt -U pip
 
+$pip install $pipopt -U $(curl -s https://raw.githubusercontent.com/ckan/ckan/$ckan_sha/requirement-setuptools.txt)
+$pip install $pipopt -r https://raw.githubusercontent.com/ckan/ckan/$ckan_sha/requirements.txt
```
ChrisBAshton, line 23: Is there a reason why this one is `-r` instead of `-U $(curl` like the others? I can't see any other examples like this in the file.

bin/run-tests.sh
```diff
@@ -1,13 +1,3 @@
...
-  coverage=" --with-coverage --cover-package=ckanext.datagovuk --cover-inclusive --cover-erase --cover-tests"
-else
-  test_name=$1
-  coverage=""
-fi
-
-# nosetests -v --nologcapture --with-pylons=${TEST_CKAN_INI:-test.ini} --with-randomly --ckan $coverage $test_name
-
-pytest --ckan-ini=test.ini $coverage $test_name
+pytest --ckan-ini=test.ini tests --cov=ckanext.datagovuk --cov-report=term-missing --disable-pytest-warnings -v
```
ChrisBAshton, line 3: Would be great if we could document these arguments in the commit message - some new ones here I think.

dev-requirements.txt
```diff
@@ -1,4 +1,3 @@
 beautifulsoup4==4.9.3
-nose-randomly==1.2.6
```
ChrisBAshton, line 2: > Also nosetests have been removed as the tests have been migrated to pytests and they have been moved in order to show test coverage.

Would be good to get this context in the commit message so that people don't have to visit GitHub to find out the reason, it's all available locally :)

Jenkinsfile
```diff
@@ -15,7 +15,7 @@ node ('!(ci-agent-4)') {
 
     stage('Installing Packages') {
       sh("rm -rf ./venv")
-      sh("virtualenv --no-site-packages ./venv")
+      sh("python3.6 -m venv ./venv")
```
ChrisBAshton, line 18: > Existing tests for ckanext-datagovuk are run on python 2.7, in order to get the tests to pass on Jenkins the venv needs to be created for python 3.

Would be great to add this context to this commit message.

Jenkinsfile
```diff
@@ -10,7 +10,7 @@ node ('!(ci-agent-4)') {
     stage('Checkout') {
       govuk.checkoutFromGitHubWithSSH(REPOSITORY)
       govuk.cleanupGit()
-      govuk.mergeMasterBranch()
+      // govuk.mergeMasterBranch()
```
ChrisBAshton, line 13: Will this line be commented out forever, or can it be removed?

ChrisBAshton (commented):
Nice one for getting to the bottom of this. It can take a lot of work to make Jenkins happy!

A few comments below, mostly about commit message structure. I also think the two "moves conf"/"moves tests" commits at the end can be squashed together into one commit, but there's no way in the GitHub UI of commenting on file renames!

bin/install-dependencies.sh
kentsanggds: I'm not sure, I just moved the code up as there was a dependency clash to do with `enum34` so I haven't actually changed how it's installed

bin/run-tests.sh
kentsanggds: OK, I'll add something to the commit message

dev-requirements.txt
kentsanggds: Will add something to commit message

Jenkinsfile
kentsanggds: Not sure what it does? Do we still need it? As doesn't github check for merge conflicts and suggests pulling things from master automatically? If we don't need to do this I'd be inclined to remove it instead

ChrisBAshton (approved):
LGTM 👍

PR #656, opened 2021-02-23
Add performance platform topics

kentsanggds:
## What

Add platform performance topics (themes) so that platform performance datasets can be tagged and easily found in the Find app.

## Reference

https://trello.com/c/NCq4ylTw/2367-3-add-topics-on-dgu-for-performance-platform

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,8 @@ node ('!(ci-agent-4)') {
     stage('Installing Packages') {
       sh("rm -rf ./venv")
       sh("virtualenv --no-site-packages ./venv")
+      sh("curl https://bootstrap.pypa.io/2.7/get-pip.py -o get-pip.py")
+      sh("bash -c 'venv/bin/python get-pip.py pip==20.3.4'")
       sh("./bin/install-dependencies.sh ./venv/bin/pip")
     }
 
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -16,8 +16,6 @@ pycsw_tag='2.4.0'
 
 pipopt='--exists-action=b'
 
-$pip install $pipopt -U pip
-
 $pip install $pipopt -U $(curl -s https://raw.githubusercontent.com/$ckan_harvest_fork/ckanext-harvest/$ckan_harvest_sha/pip-requirements.txt)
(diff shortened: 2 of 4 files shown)


bin/install-dependencies.sh
```diff
@@ -16,7 +16,8 @@ pycsw_tag='2.4.0'
 
 pipopt='--exists-action=b'
 
-$pip install $pipopt -U pip
+curl https://bootstrap.pypa.io/2.7/get-pip.py -o get-pip.py
+source venv/bin/activate && python get-pip.py pip==20.3.4
```
thomasleese, line 20: Which `venv` is this because it doesn't appear to be created in this script?

kentsanggds: it's created in the `Jenkinsfile` - https://github.com/alphagov/ckanext-datagovuk/blob/master/Jenkinsfile#L18

thomasleese: Hmm I'm wondering if the `pip` installation belongs in the `Jenkinsfile` before calling `install-dependencies.sh` because the `pip` binary is passed in as the first argument (https://github.com/alphagov/ckanext-datagovuk/blob/5347a33244e09ab25e0ff097b8f5f7e9af454baa/bin/install-dependencies.sh#L3) so it feels like that should have happened prior to this script being called.

Especially given that this script is currently generic enough that it doesn't matter what sort of virtual enviroenment you've got set up, you just need to specify which `pip` to use, but now this becomes coupled to the environment set up by the `Jenkinsfile`.

kentsanggds: sure, could move it there

Jenkinsfile
```diff
@@ -16,6 +16,8 @@ node ('!(ci-agent-4)') {
     stage('Installing Packages') {
       sh("rm -rf ./venv")
       sh("virtualenv --no-site-packages ./venv")
+      sh("curl https://bootstrap.pypa.io/2.7/get-pip.py -o get-pip.py")
+      sh("bash -c 'source venv/bin/activate && python get-pip.py pip==20.3.4'")
```
thomasleese, line 20:
```suggestion
      sh("bash -c 'venv/bin/python get-pip.py pip==20.3.4'")
```

I think this would avoid the need to run `source`.

thomasleese (approved):

PR #743, opened 2021-06-08
Only deploy main-2.9 branch to Integration

kentsanggds:
## What

Skip deployments from `main` branch which are CKAN 2.8 releases so that deployments of CKAN 2.9 can be deployed from `main-2.9` branch.

## Reference 

https://trello.com/c/GZP41t9N/2553-deploy-ckan-29-onto-integration

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,12 +40,10 @@ node ('!(ci-agent-4)') {
         }
       }
     }
-    else {
-      if (env.BRANCH_NAME == 'main-2.9') {
-        stage('Deploy CKAN 2.9 to Integration') {
-          govuk.deployIntegration('ckan', BRANCH_NAME, 'release_' + BUILD_NUMBER, 'deploy')
-        }
-      }
+    else if (env.BRANCH_NAME == 'main-2.9') {
+      stage('Deploy CKAN 2.9 to Integration') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -39,9 +39,12 @@ node ('!(ci-agent-4)') {
           govuk.dockerTagMasterBranch("ckan", env.BRANCH_NAME, env.BUILD_NUMBER)
         }
       }
-
-      stage('Deploy to Integration') {
-        govuk.deployIntegration('ckan', BRANCH_NAME, 'release_' + BUILD_NUMBER, 'deploy')
+    }
+    else {
+      if (env.BRANCH_NAME == 'main-2.9') {
```
ChrisBAshton, line 44: This can be simplified to one line:

```suggestion
    else if (env.BRANCH_NAME == 'main-2.9') {
```

ChrisBAshton (approved):
Seems reasonable to me. I don't have the full context around why you wouldn't just want to merge the 2.9 stuff into `main` and not change the deploy scripts at all, but I guess this gives you breathing room to test things out without potentially putting buggy code on `main`.

Jenkinsfile
kentsanggds: Thanks, wasn't sure if this was possible, will try that out

PR #818, opened 2021-08-25
Only allow organogram xls

kentsanggds:
## What

Only allow publishers to upload XLS files, this should help reduce problems with Organogram CSV files.

## Reference 

https://trello.com/c/UGpKiiOb/2673-some-organograms-dont-render-their-interactive-diagrams


ChrisBAshton (dismissed):
LGTM 👍

PR #848, opened 2021-09-15
Fix users password reset by relaxing the username schema

kentsanggds:
## What

Users are unable to change their passwords because their Usernames are not valid when using the default CKAN user schema, so relax the user schema by monkey patching it.

Also prevent organogram errors from reaching sentry as its an error that can be handled by the user and fix a bug in making the organogram filenames lowercase.

## Reference 

https://trello.com/c/alHbD02j/2705-users-cannot-change-their-ckan-password

### DIFF
--- a/ckanext/datagovuk/ckan_patches/cli.py
+++ b/ckanext/datagovuk/ckan_patches/cli.py
@@ -16,3 +16,5 @@ def set_password(ctx, username):
 
     _user = model.User.get(username)
     send_password_alert(_user)
+
+ckan.cli.user.set_password = set_password
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -294,6 +294,7 @@ class DatagovukPlugin(plugins.SingletonPlugin, toolkit.DefaultDatasetForm, Defau
         "is not a valid format",                                    # Harvest
         "Gather stage failed",                                      # Harvest
         "Errors found by ETL were not picked up by spreadsheet",    # Datagovuk
+        "Organogram template XLS file expected but got:",           # Datagovuk
         "User not found",                                           # CKAN
         "Group not found",                                          # CKAN
         "not authorised to",                                        # CKAN
--- a/ckanext/datagovuk/plugin.py
+++ b/ckanext/datagovuk/plugin.py
@@ -220,6 +220,7 @@ class DatagovukPlugin(plugins.SingletonPlugin, toolkit.DefaultDatasetForm, Defau
             DGUUserRegisterView,
             me,
         )
+        from ckanext.datagovuk.schema import user_new_form_schema, user_edit_form_schema
 
         bp = Blueprint("datagovuk", self.__module__)
         bp.template_folder = u'templates'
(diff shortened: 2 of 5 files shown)


ckanext/datagovuk/plugin.py
```diff
@@ -257,10 +253,15 @@ def dgu_home():
...
         # call it directly instead of redirecting externally
         import ckan.views.user as ckan_user_views
+        ckan_user_views._edit_form_to_db_schema = user_edit_form_schema
+        ckan_user_views._new_form_to_db_schema = user_new_form_schema
         ckan_user_views.me = me
 
         return bp
 
+    # import these for monkey patching
+    from ckanext.datagovuk.ckan_patches import cli, logic, query
```
richardTowers, line 263: It feels dangerous to have the monkey patching happen as a side effect of an import. Do you actually need `logic` and `query` here too?

Was there something wrong with the old approach? Or could we have a separate function on `ckan_patches` which does the monkey patching like:

```
from ckanext.datagovuk.ckan_patches.cli import monkey_patch
monkey_patch()
```

kentsanggds: The old approach included these monkey patches. Logic is actually the one that handles the user schema, query blocks some search queries which were causing problems in the past. 
This has been fixed by upgrading the service affected but it's probably a good idea to continue blocking the tags.

richardTowers: Before this PR, `logic` and `query` weren't being imported, or am I missing something? Are you saying that they used to be imported, before the upgrade, but they got removed at some point? So now we're adding them back?

I see the approach of monkey patching as a side effect of an import has been around for the last 5 years (in `logic.py`), so I guess I'm fine with it for this PR. It's a big code smell though - imports shouldn't have side effects, and they doubly shouldn't have monkey patching side effects.

ckanext/datagovuk/ckan_patches/cli.py
```diff
@@ -16,3 +16,5 @@ def set_password(ctx, username):
 
     _user = model.User.get(username)
     send_password_alert(_user)
+
+ckan.cli.user.set_password = set_password
```
huwd, line 20: What is the significance of this line moving from plugin.py to here in ckan_patches/cli.py?

ckanext/datagovuk/plugin.py
kentsanggds: Pre 2.9 they were being imported, it was removed in this commit https://github.com/alphagov/ckanext-datagovuk/commit/270bc2da195cbc46ca38f2afcc0eec1d44da7647#diff-b5ec20d532632efd2db909f73caa5ff88a59ab0a1664dd6ceb3a8f4a31ae4df9 probably by accident as that commit was part of the 2.9 upgrade branch. 
I think the problem was that there weren't any tests in place to check that it was allowing through more relaxed user names as the user schema in DGU is much more relaxed, this test has now been added in to prevent this from happening in the future.

ckanext/datagovuk/upload.py
```diff
@@ -77,7 +77,7 @@ def upload_resource_to_s3(context, resource):
     filename = (
         resource.get("timestamp", timestamp.strftime("%Y-%m-%dT%H-%M-%SZ"))
         + "-"
-        + slugify(resource.get("name"), lowercase=True)
+        + slugify(resource.get("name")).lower()
```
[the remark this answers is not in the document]
kentsanggds: Before we were using `awsome-slugify` which doesn't appear to work beyond python 3.4, this was only picked up after doing a `force-reinstall`, so I moved towards using `python-sulgify`, unfortunately `lowercase=True` is not the same as `to_lower=True` so I just used the standard `lower()` to ensure that the filename is lowercased.

ckanext/datagovuk/ckan_patches/cli.py
kentsanggds: This keeps it more in line with the other patches in the `ckan_patches` directory, and also means that it can just be imported in `plugin.py`

richardTowers (approved):
Changes make sense now they've been explained.

I still think the pattern of monkey patching as a side effect of an import is extremely suspect, but it's been like that for many years now so it's not worth blocking this PR on that.

huwd:
You can dismiss my questions and go with Richard's ✅  if you want,
I might still need insight into what's going on in those places before I feel informed enough to approve!

PR #917, opened 2021-12-14
Add banner database upgrade

kentsanggds:
## What 

Show banner to users with dates for downtime due to database upgrade.

## Reference

https://trello.com/c/lipeZ92R/50-prepare-comms-for-ckan-downtime


richardTowers (approved):
Looks good to me

PR #970, opened 2022-02-23
Update ckan spatial dependency

KludgeKML:
Updates to use latest version of alphagov/ckanext-spatial, with validation fixes for Gemini 2.3 Schemas.

### DIFF
--- a/bin/install-dependencies.sh
+++ b/bin/install-dependencies.sh
@@ -9,7 +9,7 @@ ckan_dcat_fork='alphagov'
 ckan_dcat_sha='be3b809fa7431c8d81508c01853e81ce6a5dfd84'
 
 ckan_spatial_fork='alphagov'
-ckan_spatial_sha='3f423de1e9e4e6975725712626d904887779b408'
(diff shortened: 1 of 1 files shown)


bin/install-dependencies.sh
```diff
@@ -9,7 +9,7 @@ ckan_dcat_fork='alphagov'
 ckan_dcat_sha='be3b809fa7431c8d81508c01853e81ce6a5dfd84'
 
 ckan_spatial_fork='alphagov'
-ckan_spatial_sha='3f423de1e9e4e6975725712626d904887779b408'
+ckan_spatial_sha='2a21094b7e63ee48b8beb1229997827b66a62fe4'
```
kentsanggds, line 12: Should it be -

```suggestion
ckan_spatial_sha='433c7f2a4ba0b05871fe8bbc9b0518aa594f392d'
```

as grabbing the commit from `main-2.9` branch

KludgeKML: Lol, caught myself out with branches again. Yes, that's what it should be. I'll commit the suggestion.

PR #1178, opened 2022-11-02
Remove Jenkins Docker build tasks

kevindew:
Trello: https://trello.com/c/amnaU0V9/853-decommission-the-publishing-end-to-end-tests

The Docker functions are being removed from the govuk-jenkinslib as
their most significant need, publishing-e2e-tests, has been retired [1].

As far as I can tell no production systems make use of these images.
When this project is used by replatforming these will need to be pushed
to a different destination (ECR).

[1]: https://github.com/alphagov/govuk-jenkinslib/pull/120

This also resolves a deprecated function.


kentsanggds (approved):
LGTM

PR #1219, opened 2022-12-22
Update ci.yaml to use standard CKAN solr image

kentsanggds:
## What

Reinstate the github action tests to run so that other orgs can submit changes to things like the organograms and see automated test results.


kevindew (approved):
Seems good to me based on my very meagre understanding.

PR #1260, opened 2023-02-10
Fix docker

kentsanggds:
## What

A number of changes have been made to accommodate deployment to a k8s cluster - 

- Update the upload code to check for access key and secrets and if they are set.
- Add pycsw dockerfile and update the docker-compose file to use it
- Create a default user to enable visual regression tests to pass
- update the spatial sha to one updated to work with pycsw
- use an updated production.ini to make it work better within a docker stack

## Reference 

https://trello.com/c/CxQutDOt/1115-update-ckan-helm-chart-to-run-on-integration-cluster
https://trello.com/c/uWgUDRoc/1103-fix-ckan-visual-regression-tests-to-run-against-297


samsimpson1 (approved):
All good as long as `Dockerfile.pycsw` is renamed to `pycsw.Dockerfile`

PR #1279, opened 2023-03-01
Add cmd reindex recent datasets

kentsanggds:
## What

Add a CLI command to just reindex recent datasets, default is set to 24 hours before now, otherwise can set an env var to adjust the timeframe. 

Also update the solr_reindex_target_date script to handle reindexing of datasets with different states and add the DB script to set duplicate datasets to a deleted state where they have a harvest object which is not the current one.

## Reference 

https://trello.com/c/5LSihxlF/1139-investigate-and-fix-duplicate-datasets
https://trello.com/c/CxQutDOt/1115-update-ckan-helm-chart-to-run-on-integration-cluster

### DIFF
--- a/bin/python_scripts/solr_reindex_target_date.py
+++ b/bin/python_scripts/solr_reindex_target_date.py
@@ -5,7 +5,7 @@
 # export POSTGRES_URL=<sqlalchemy.url from ckan.ini>
 #
 # Execute script like this -
-# python solr_reindex_target_date.py yyyy-mm-dd <optional command: show, reindex>
+# python solr_reindex_target_date.py yyyy-mm-dd <optional command: show, reindex> <optional state: (default is active)>
 #
 
 import os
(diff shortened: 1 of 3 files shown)


bin/python_scripts/solr_reindex_target_date.py
```diff
@@ -40,22 +40,22 @@ def is_dev():
...
 
 
-def get_datasets_with_resource_target_date(target_date):
+def get_datasets_with_resource_target_date(target_date, state):
     cursor = connection.cursor()
     sql = """
-    SELECT name FROM package WHERE id IN (SELECT package_id FROM resource WHERE last_modified = '%s')  and state='active';
-    """ % (target_date)
+    SELECT name FROM package WHERE id IN (SELECT package_id FROM resource WHERE last_modified = '%s')  and state='%s';
+    """ % (target_date, sate)
```
beccapearce, line 58: I think this should say state?

kentsanggds: ah yes, I'll correct it

beccapearce (approved):

PR #1313, opened 2023-03-30
Add 2.9.3 Dockerfile for CKAN

kentsanggds:
This will make it easier to develop the stack in line with the production version of CKAN.

### DIFF
--- a/docker/ckan/2.9.3.Dockerfile
+++ b/docker/ckan/2.9.3.Dockerfile
@@ -52,7 +52,7 @@ RUN mkdir -p $CKAN_VENV $CKAN_CONFIG $CKAN_STORAGE_PATH && \
 ENV PATH=${CKAN_VENV}/bin:${PATH}  
 
 # copy production.ini
-ADD . $CKAN_VENV/src/ckanext-datagovuk/
+COPY . $CKAN_VENV/src/ckanext-datagovuk/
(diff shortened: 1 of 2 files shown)


docker/ckan/2.9.3.Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+# Setup virtual environment for CKAN
+RUN mkdir -p $CKAN_VENV $CKAN_CONFIG $CKAN_STORAGE_PATH && \
+    python3.6 -m venv $CKAN_VENV && \
+    ln -s $CKAN_VENV/bin/ckan /usr/local/bin/ckan
+
+# Virtual environment binaries/scripts to be used first
+ENV PATH=${CKAN_VENV}/bin:${PATH}  
+
+# copy production.ini
+ADD . $CKAN_VENV/src/ckanext-datagovuk/
```
sengi, line 55: For copying from the host or another container (as opposed to downloading a URL), `COPY` is preferable to `ADD` because it has fewer surprising features / gotchas.

sengi (approved):

docker/ckan/2.9.3.Dockerfile
kentsanggds: ok, will fix that

PR #1351, opened 2023-05-09
Disable building images on schedule

samsimpson1:
This functionality doesn't work since the build needs extra parameters that this trigger doesn't support, so it always fails.


kentsanggds (approved):
Will look at the build failures later after ruby upgrades

PR #1383, opened 2023-06-15
Upgrade ckan 2.10

kentsanggds:
## What

Upgrade of CKAN to 2.10, this will be merged into main-2.10 after approval as the existing govuk stack is still using 2.9.3 and won't be able to support 2.10 until CKAN is replatformed.

## Reference

https://trello.com/c/bEsbEk1m/1151-run-security-check-and-resolve-critical-issues


ckan.ini
```diff
@@ -0,0 +1,338 @@
...
+ckan.auth.enable_cookie_auth_in_api = true
+ckan.auth.route_after_login = dashboard.datasets
+
+## User Account Creation Setting
+; ckan.valid_email_regexes = .gov.uk$ .nhs.uk$ .nhs.net$ .ac.uk$ .os.uk$ .mod.uk$ .police.uk$ .bl.uk$
+
+## CSRF Protection #############################################################
+WTF_CSRF_ENABLED = true
+WTF_CSRF_CHECK_DEFAULT = true
+WTF_CSRF_SECRET_KEY = _e6qkoYldyll9U3zfy5n6rg-JxCPY4DUfy5LHu3qnUo
```
sengi, line 114: Is this used for anything real? If so you probably need to change it 😅

sengi (approved):
All looks reasonable to me, apart from the suspicious-looking secret string in the config file that you might need to change (unless it's unused, in which case I'd remove it or change it to something that's obviously a test/dummy value, whatever makes sense for your use case)

ckan.ini
kentsanggds: Not used for anything real, just auto generated so should be ok for local dev use, I'll make sure that the ckan chart auto generates when deployed

kentsanggds:
Closing this in favour of https://github.com/alphagov/ckanext-datagovuk/pull/1662

PR #1438, opened 2023-08-04
Set the image tag to release tag on tag trigger

kentsanggds:


docker/create-pr.sh
```diff
@@ -2,9 +2,14 @@
 
 set -eux
 
-export IMAGE_TAG=$(gh api repos/alphagov/ckanext-datagovuk/branches/${GH_REF} | jq .commit.sha -r)
+if [[ ${IS_TAG:-} = "true" ]]; then
+  export IMAGE_TAG=${GH_REF}
```
sengi, line 6: nit: shouldn't matter in this case, but slightly safer to write:
```suggestion
  export IMAGE_TAG="${GH_REF}"
```

sengi (approved):

PR #1442, opened 2023-08-07
Fix create pr add build image tags

kentsanggds:
- Fix the create PR so that the branch name is used for checking out github and creating the PR
- also rename ghcr to build-image to make it clearer 
- and add in a build image for the tag version so that images can be quickly identified to a tagged release

### DIFF
--- a/.github/workflows/build-image-on-tags.yaml
+++ b/.github/workflows/build-image-on-tags.yaml
@@ -1,0 +1,42 @@
+name: Build and push images
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  build_and_push:
+    name: ${{ matrix.app.name }} (${{ matrix.arch }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        app:
+          - name: ckan
+            version: 2.9.9
+          - name: pycsw
+            version: 2.6.1
+          - name: postgis
+            version: 13-3.1
+          - name: solr
+            version: 8
+        arch: [ amd64 ]
+    permissions:
+      packages: write
+    steps:
+      - name: Login to GHCR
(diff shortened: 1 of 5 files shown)


.github/workflows/build-image-on-tags.yaml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  build_and_push:
+    name: ${{ matrix.app.name }} (${{ matrix.arch }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
sengi, line 13: Unusual use of matrix builds, but no probs :)

docker/create-pr.sh
```diff
@@ -29,6 +30,6 @@ for ENV in $(echo $ENVS | tr "," " "); do
     done
     git commit -m "Update image tags for ${ENV} to ${IMAGE_TAG}"
     git push --set-upstream origin "${BRANCH}"
-    gh pr create --title "Update image tags for ${ENV} (${IMAGE_TAG})" --base main --head "${BRANCH}-${ENV}" --fill
+    gh pr create --title "Update image tags for ${ENV} (${IMAGE_TAG})" --base main --head "${BRANCH}" --fill
```
sengi, line 33: You might find the PR workflow for automated commits a little tricky. For the rest of GOV.UK we just commit and merge directly. No problem if you wanna experiment with this though.

sengi (approved):
I suspect there are actions and/or workflows you could reuse from [govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure/tree/main/.github).

I should have thought to suggest this earlier (my bad) — but worth chatting with @theseanything about the whole CI/CD / build/release automation thing as he might have suggestions that could save you a lot of hassle.

docker/create-pr.sh
kentsanggds: I think lets try it out first, it will probably be mainly me who will be running the process

PR #1461, opened 2023-08-25
Update dockerfile spatial gemini 2.3

kentsanggds:
This PR contains a number of updates - 

- Updates the 2.9.9-base.Dockerfile to use a spatial sha that works with the gemini-csw harvest source.
- Fix the Dockerfiles as they would not have picked up changes in this repo in their previous form.
- Update the build-image.yaml to allow for clearer options for building and pushing images
- Make fixes to the docker-compose-2.9.3.yml in line with updated directory structure

### DIFF
--- a/docker/ckan/2.9.9-base.Dockerfile
+++ b/docker/ckan/2.9.9-base.Dockerfile
@@ -1,6 +1,8 @@
 FROM ghcr.io/alphagov/ckan:2.9.9-core
 
 COPY production.ini $CKAN_CONFIG/production.ini
+# Set CKAN_INI
+ENV CKAN_INI $CKAN_CONFIG/production.ini
 
 RUN chown -R ckan:ckan $CKAN_HOME $CKAN_VENV $CKAN_CONFIG $CKAN_STORAGE_PATH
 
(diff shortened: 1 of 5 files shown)


docker/ckan/2.9.9-base.Dockerfile
```diff
@@ -1,16 +1,10 @@
...
-    cp -v $CKAN_VENV/src/ckanext-datagovuk/bin/setup_ckan.sh /ckan-entrypoint.sh && \
-    chmod +x /ckan-entrypoint.sh && \
-    chown -R ckan:ckan $CKAN_HOME $CKAN_VENV $CKAN_CONFIG $CKAN_STORAGE_PATH
-
+COPY production.ini $CKAN_CONFIG/production.ini
 # Set CKAN_INI
 ENV CKAN_INI $CKAN_CONFIG/production.ini
 
-WORKDIR $CKAN_VENV/src/ckanext-datagovuk/
+RUN chown -R ckan:ckan $CKAN_HOME $CKAN_VENV $CKAN_CONFIG $CKAN_STORAGE_PATH
```
sengi, line 7: Not a problem for now, but I suspect the app doesn't really need to write to most of these. Something to revisit after launch maybe.

sengi (approved):

docker/ckan/2.9.9-base.Dockerfile
kentsanggds: Sure, I'll have a look at this after sorting out all the launch stuff, I think most of this came from the official CKAN Dockerimage

PR #1482, opened 2023-09-12
Fix solr version in build-image-on-tags.yaml

kentsanggds:


sengi (approved):
Whoops, [Mariner 1](https://www.wired.com/2009/07/dayintech-0722/) vibes there 😅

PR #1509, opened 2023-10-03
Update create PR on tags on workflow trigger

kentsanggds:
Rather than creating the PR for staging and production on tag push, wait for the image to be built before creating the PRs


dj-maisy (approved):
👍

PR #1699, opened 2024-04-29
Pin lxml to < 5.1.1

kentsanggds:
5.1.1 removes a require attribute _ElementStringResult which is causing problems with some harvest sources using XML.

### DIFF
--- a/docker/ckan/2.9.9.Dockerfile
+++ b/docker/ckan/2.9.9.Dockerfile
@@ -19,7 +19,10 @@ RUN echo "pip install ckanext-datagovuk..." && \
 
     # install ckanext-datagovuk
     pip install $pipopt -U -r requirements.txt && \
-    pip install $pipopt -U -e . 
+    pip install $pipopt -U -e . && \
+
+    # pin lxml as 5.1.1 removes required property _ElementStringResult
+    pip install lxml==5.1.0
 
 # to run the CKAN wsgi set the WORKDIR to CKAN
 WORKDIR "$CKAN_VENV/src/ckan/"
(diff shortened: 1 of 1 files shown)


sengi (commented):
Would be less confusing/brittle to change the existing constraint in [requirements.txt](https://github.com/alphagov/ckanext-datagovuk/blob/db8c6b8/requirements.txt#L6) to `<5.1.1` (with a TODO for the underlying issue)?

kentsanggds:
I could change it to that, not sure what the TODO would be as the upgrade to 2.10 might fix it anyways as CKAN have made a bunch of changes to ckanext-spatial which is the extension that actually uses lxml. Maybe the TODO could be just to revisit the pin for 2.10?

docker/ckan/2.9.9.Dockerfile
```diff
@@ -19,7 +19,11 @@ RUN echo "pip install ckanext-datagovuk..." && \
 
     # install ckanext-datagovuk
     pip install $pipopt -U -r requirements.txt && \
-    pip install $pipopt -U -e . 
+    pip install $pipopt -U -e . && \
+
+    # pin lxml as 5.1.1 removes required property _ElementStringResult
+    # TODO: revisit for CKAN 2.10 upgrade as might not be required
+    pip install lxml<5.1.1
```
sengi, line 26: Did you mean to do this in requirements.txt? Seems confusing to be overriding it here.

kentsanggds: Could do it in the ckanext-spatial requirements but also need to investigate whether the upgrade to 2.10 fixes this issue which might delay this fix. So I'm ok with just pinning it here for now

sengi: No worries 👍

PR #1777, opened 2024-07-15
Update Solr, Add --target to Dockerfile FROM layers

dj-maisy:
## What?
This adds the `--platform` flag to each of the Dockerfiles to prepare us for multi-arch builds.

This also changes the Solar Dockerfile so we're cutting out the `ckan/ckan-solr:2.10-solr8-spatial` Base image, CKAN don't build this for ARM, but as it is based on Solr's own images that _are_ already built for ARM, we can just cut out the "middle man" so to speak and build it ourselves.


docker/ckan/2.10.3-core.Dockerfile
```diff
@@ -1,6 +1,6 @@
 # See CKAN docs on installation from Docker Compose on usage
 # FROM ubuntu:focal
-FROM ubuntu:jammy
+FROM --platform=$TARGETPLATFORM ubuntu:jammy
```
kentsanggds, line 3: We can probably remove this file as we have upgraded to 2.10.4

docker/ckan/2.10.3.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/alphagov/ckan:2.10.3-base
+FROM --platform=$TARGETPLATFORM ghcr.io/alphagov/ckan:2.10.3-base
```
kentsanggds, line 1: This file can be removed as upgraded to 2.10.4

docker/ckan/2.9.9-base.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/alphagov/ckan:2.9.9-core
+FROM --platform=$TARGETPLATFORM ghcr.io/alphagov/ckan:2.9.9-core
```
kentsanggds, line 1: This file can be removed as upgraded to 2.10.4

docker/ckan/2.9.9-core.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # See CKAN docs on installation from Docker Compose on usage
-FROM ubuntu:focal
+FROM --platform=$TARGETPLATFORM ubuntu:focal
```
kentsanggds, line 2: This file can be removed as upgraded to 2.10.4

docker/nginx/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nginx:stable-alpine
+FROM --platform=$TARGETPLATFORM nginx:stable-alpine
```
kentsanggds, line 1: This is only used locally so probably don't need to update?

docker/solr/2.10.Dockerfile
```diff
@@ -1,4 +1,58 @@
-FROM ckan/ckan-solr:2.10-solr8-spatial
+FROM --platform=$TARGETPLATFORM solr:8
```
kentsanggds, line 1: I'll have to test it out locally to see if this still works with CKAN 2.10.4, and then we will probably have to update the patch version for solr in the build image GHA

kentsanggds (changes requested):
A few changes as some files are not needed so probably best removed

kentsanggds (approved):

samsimpson1 (approved):

dj-maisy:
I'm happy for this to be merged and then we can look at sorting out the Workflows.

kentsanggds:
Unfortunately once this is merged in it will affect the building of the base images, so I'm going to add some more commits to handle that in another branch to merge into this branch before merging to main

PR #1785, opened 2024-07-24
Use df.loc[index, col] as df[col, index] is being deprecated

kentsanggds:
This fixes an issue with organogram uploads reported by HMRC


AgaDufrat (approved):
Had a look at [the `pandas.DataFrame.loc` doc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html) and LGTM.

PR #1792, opened 2024-07-30
Add PATCH env to build-image-on-tags.yaml

kentsanggds:
Without the PATCH env arg the build image won't be correct when releasing to Staging and Production.


theseanything (approved):

dj-maisy (approved):
👍

Issue #1797, opened 2024-07-31
Refactor Build Workflow for Multi-Arch Support


dj-maisy:
## What?
We've updated the Dockerfiles to be able to support ARM builds, however, the current Workflows are not set up to be able to take advantage of this.

We should refactor the "Build" Workflows in a similar fashion to the [shared GOV.UK Workflows](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/build-and-push-multiarch-image.yml).

PR #1820, opened 2024-08-28
Refactor Build Workflow for ARM Build Support

dj-maisy:
## Preface
This took quite a bit of work! And sitting and thinking about "how" to make this somewhat maintainable, understandable and also not too repetitive/redundant. Always happy for suggestions, recommendations, constructive criticisms etc.

## What?
This creates a new Workflow file for CKAN/DGU builds to create ARM versions of each of the different images. The new workflow tries to do-away with the current `build-image.sh` bash script and move a lot of the logic into the Workflow itself. The hope is that this will make the workflow a little easier to debug if anything goes wrong.

This PR also creates a new `build-config.yaml` file that contains the various app versions, tags and also the "build types" and which apps each of those build types will actually build, using YAML anchors to try and de-duplicate as much of it as possible. This is a slight deviation from the JSON config file we used for the GOV.UK Ruby Bases, which had a lot less to duplicate. This file is also where we put the runner type configuration.

## Related

* Resolves https://github.com/alphagov/ckanext-datagovuk/issues/1797
* Related https://github.com/alphagov/govuk-helm-charts/issues/2165

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -108,6 +108,7 @@ jobs:
               echo "BUILD_TAGS=${{ matrix.app.version }}" >> $GITHUB_ENV
               echo "DOCKERFILE=${{ matrix.app.version }}" >> $GITHUB_ENV
               echo "BUILD_CKAN_BASE=true" >> $GITHUB_ENV
+              echo "ADD_PATCH_TAG=true" >> $GITHUB_ENV
               echo "APP_TAG_SUFFIX=-core" >> $GITHUB_ENV
             ;;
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build-image.yaml
```diff
@@ -1,109 +1,365 @@
...
+        description: Decide on what to build
+        required: true
+        type: string
+      gitRef:
+        required: false
+        type: string
+        default: ${{ github.sha }}
   workflow_dispatch:
     inputs:
       buildType:
```
kentsanggds, line 16: Previously the build types were in `build-config.yaml`, is that file now redundant?

dj-maisy: The file is still necessary to define the various app version and patch IDs. It also defines which apps are built by which "build types". So we do need to keep it.

build-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+    version: "2.10.4"
+    patch: e
+  pycsw: &app_pycsw
+    name: pycsw
+    version: "2.6.1"
+    patch: k
+  solr: &app_solr
+    name: solr
+    version: "2.10"
+    patch: a
```
kentsanggds, line 13: This will probably need bumping up to `b` so that we can also build it using the same build process and for testing the process now.

kentsanggds (approved):
I did some testing on integration and it looks like its working as expected, just need to tidy up the commits now and then its ready to merge in

dj-maisy:
Done an interactive rebase and squashed most of the surplus commits.

PR #1825, opened 2024-09-04
Upgrade python 3.11

kentsanggds:
Upgrading to python 3.11 should fix iso date format validation issues for fractional seconds reported by a publisher.


KludgeKML (approved):
Looks okay to me

PR #1851, opened 2024-10-04
Add SQL script for reseting updated_at value for OS roads

kentsanggds:
This will ensure that the latest harvest_object_id is set.

The script will have to be run manually as there is no automatic application of these database scripts.


AgaDufrat (approved):
LGTM. Do we know why `harvest_object_id` wasn't updated and have we now got something in place to prevent this happening in the future?

kentsanggds:
It's not really worth investigating it as the plan is to remove Publish from the DGU tech stack completely.

PR #1859, opened 2024-10-24
Remove PVA duplicates

kentsanggds:
Rename script to remove_duplcates.py

Also update it for python 3 and add in SQL script to remove PVA duplicates


kentsanggds:
Redundant as more duplicates were found and fixed in this PR https://github.com/alphagov/ckanext-datagovuk/pull/1879

PR #1905, opened 2024-12-18
Bump Solr to 2.10-c for ARM flavour

dj-maisy:
## What?
This just bumps the patch version for Solr to "c" so we can have a tagged ARM version available for testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       options: --user root
     services:
       solr:
-        image: ghcr.io/alphagov/solr:2.10-b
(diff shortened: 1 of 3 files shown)


AgaDufrat (approved):
Seems ok to me.
@kentsanggds were just talking about upgrading Solr. I thought we are using Solr 8 https://github.com/alphagov/ckanext-datagovuk/blob/main/docker/solr/2.10.Dockerfile#L1

kentsanggds (changes requested):
Please fix the warnings in the solr/2.10.Dockerfile, thanks!

kentsanggds (approved):
LGTM, the solr image might need to be built manually to pass the CI checks

PR #1948, opened 2025-02-11
⬆️ Bump actions/cache to 4.2.0

jasonBirchall:
Actions cache v2 has started to brownout. We should look to move to the latest version.


nimalank7 (approved):
LGTM!

PR #1950, opened 2025-02-11
✨ Don't trigger CI when making doc changes

jasonBirchall:
This commit stops CI from runnning if only markdown files have changed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
       - main
     paths-ignore:
       - "**.md"
+      - ".gitignore"
+      - ".github"
+      - ".dockerignore"
     types:
       - closed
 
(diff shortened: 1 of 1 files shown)


nimalank7:
I think we also want to exclude `.gitignore`, `.dockerignore`, `.github` etc... as well

nimalank7:
Just spotted the `LICENSE` file as well.

jasonBirchall (commented):
> Just spotted the LICENSE file as well.

I don't want comprehensive coverage here :)
I appreciate the review, but we should make additions in another PR

AgaDufrat (approved):

AgaDufrat:
> Just spotted the LICENSE file as well.

LICENSE gets deployed so I'm sure if we want to exclude it. We should probably rename it to `LICENCE` for consistency.

PR #2018, opened 2025-04-29
Catch value error organograms

kentsanggds:
Some publishers were having problems with the organograms but not able to see the all the validation errors. 
This should catch all ValueErrors to report back to the publisher so that they can amend their organogram file.

This has been tested on integration and works as expected.


nicholsj (approved):
Thanks!

PR #2067, opened 2026-01-07
Update local docker compose stack

kentsanggds:
Add in Find to the stack and introduce a Makefile to make it easier to run commands.

The README was substantially updated as there has been significant changes in the easiest way to get it running locally.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/362?selectedIssue=DGUK-50

### DIFF
--- a/docker/bootstrap.sh
+++ b/docker/bootstrap.sh
@@ -2,26 +2,16 @@
 
 # Default git fork
 CKAN_FORK=ckan
-CKAN_SHA=70bec6611003b96ec7a02abb9605afdfb0964ad9
+CKAN_SHA=9f2094b26721b28918d45d58fea16176df7a1ce1
 DCAT_FORK=ckan
 DCAT_SHA=618928be5a211babafc45103a72b6aab4642e964
-SPATIAL_FORK=alphagov
-SPATIAL_SHA=de1f67cb7535d9b73079be6da56bf1b5d919a641
-HARVEST_FORK=ckan
-HARVEST_SHA=cb0a7034410f217b2274585cb61783582832c8d5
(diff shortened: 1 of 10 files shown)


docker/bootstrap.sh
```diff
@@ -2,26 +2,16 @@
...
-    SPATIAL_FORK=alphagov
-    SPATIAL_SHA=d64fd925700990a352ef23ed180c47dbc7277138
-    CKAN_SHA=885f9e0b668e3496a8f2c0c0a9f1cb59bf810e16
-    CKAN_FORK=alphagov
-    SRC_DIR=2.9.3
-    DCAT_FORK=ckan
-    DCAT_SHA=d64fd925700990a352ef23ed180c47dbc7277138
-fi
+HARVEST_SHA=9fb44f79809a1c04dfeb0e1ca2540c5ff3cacef4
+SRC_DIR=2.10.7
```
ashimali: Ha, yes I tried at end of day to run the bootstrap followed by the make target and got a dir version number error as it does fresh checkout of harvest. So we'll have an ongoing version number dependency in bootstrap process?

ashimali (approved):
Looks good to me, and thanks for adding the bootstrap make (and note to readme as well as I only found that in the echo output from boostrap.sh)

docker/bootstrap.sh
ashimali: Actually still a few version errors here are there. Still references to 2.10 rather than 2.10.7

kentsanggds: Lets change the SRC_DIR to 2.10, as this will probably be the least effort

PR #2069, opened 2026-02-26
Fix smtp server connection

kentsanggds:
Use the forked version of CKAN to fix an issue with the SMTP server connections due to Amazon adopting stricter stricter SNI (Server Name Indication) validation.

Also had to pin setuptools to 80 as there is a breaking change for versions > 80 which affect the core CKAN code.

Finally dropped the build of the AMD64 image as we are now running ARM64 images on the EKS cluster.

### DIFF
--- a/build-config.yaml
+++ b/build-config.yaml
@@ -2,7 +2,7 @@ apps:
   ckan: &app_ckan
     name: ckan
     version: "2.10.7"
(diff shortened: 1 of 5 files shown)


ashimali:
👍🏼

ashimali (approved):

PR #2072, opened 2026-04-20
feat: This branch is intended to slightly rationalise bootstrap process

ashimali:
for local dev.

Update bootstrap process to rely less on files with version in file name.

Move CKAN_VERSION number to .env.example and have bootstrap.sh copy to .env

Remove old checked out ckan dirs left behind in gitignored docker/src dir.

bootstrap can be run again with already checkout out docker/src dependency dir but asks user to confirm before destructive change.

Not in this PR: reducing docker/ckan Dockerfiles to core.Dockerfile, base.Dockerfile and dev.Dokerfile and Dockerfile - therefore docker/ckan Dockerfiles still embed version number in file name.


Obsiye:
looks good to me. I did see a file called docker/start-ckan.sh which i think would need updating too or deleting.

ashimali:
@kentsanggds can you take a look and see if we can take this further and clean up all version numbered files. What impact if any does this have for anything downstream, given that in theory this is a local dev concern?

PR #2076, opened 2026-04-23
DGUK-68-feat/check links

ashimali:
This PR has two scripts 

1. check_links.py
 
Checks active resource URLs of active dataset packages.
For each URL:
  * Classifies the response (OK / 404 / 410 / other 4xx / 5xx / timeout / etc.).
  * in `dry-run` mode no db writes - just report/log output
  * in `live' mode: if 404 or 410, updates resource `state` to 'deleted'
  * and updates package `metadata_modified` to NOW(). - not 100% sure this is needed since we pass reosource ids to reindex?
  * added today 28/04/2026 - a `limit` number  argument that might be useful for sanity check run before letting loose on whole db in one go. This will mean script will process the number of resources passed in as an arg.
Writes 
`check_links_report_{ts}.csv` 
`packages_to_reindex_{ts}.txt`

The packages to reindex file is for processing by:

 `solr_reindex_package_ids.py`.

2. revert_links_deletion.py

Reverses a prior run of `check_links.py --mode live` run using its CSV output report.
For every row with `to-delete == "true"`:
  * `live`: flips the resource `state` back to 'active' (if currently
    'deleted') and bumps the package `metadata_modified` to NOW().
  * `dry-run`: logs the intended change, no DB writes.

Also writes a `packages_to_reindex_revert_{ts}.txt` for `solr_reindex_package_ids.py`.


bin/python_scripts/check_links.py
```diff
@@ -0,0 +1,417 @@
...
+    "category",
+    "error-detail",
+    "to-delete",
+    "checked-at",
+]
+
+# TODOs:
+#  - Add org to csv output.
+#  - Add more logging to this script covering steps
+#  - rate limiting - let's see if we get limited and introduce some throttling
```
joelanman: I think we should probably do rate limiting - its polite and some of these catalogues/servers are a bit ropey

ashimali: 👍🏼

bin/python_scripts/script-runbook.md
```diff
@@ -0,0 +1,61 @@
...
+
+Set environment variable for: `POSTGRES_URL`
+
+```bash
+    python check_links.py --mode dry-run --workers 50
+```
+
+CSV output will show results, including column `to-delete` which which show resources that would be updated in live mode.
+
+To update db instead of `--mode dry-run` use `--mode live`
```
Obsiye: i'm not sure what the database backup/snapshot configuration is like.

should this mention "Make sure to take a manual snapshot of the db" ? or something like that

ashimali: Good point. I don't actually know what the backup strategy is, although I know there is a dump in S3, I don't know the regularity of of. On the other hand, the ckan convention is you don't actually delete, but set state = 'deleted' so we can reverse the process.

bin/python_scripts/check_links.py
ashimali: So I've introduced two things short of plugging in something like requests-ratelimiter with host based rate limits. 

for now how about we try the following first and to see how it goes:

- Added in a std lib retry with back off  that is configurable 
- And on the other side, the data coming out of the db, I added a round robin interleave of hosts, so that the urls are more distributed across the result set so that we don't get as many clusters of host domains the the workload.

bin/python_scripts/check_links.py
```diff
@@ -0,0 +1,420 @@
...
+    "resource-url",
+    "http-status",
+    "category",
+    "error-detail",
+    "to-delete",
+    "checked-at",
+]
+
+# TODO:
+#  Add org to csv output - not sure how to do this without having
```
kentsanggds: This might be useful if we intend to reach out to orgs to check on some of the broken links. 

Also will the report group resources belonging to a dataset together and also group datasets under an org?

ashimali: Yes it would be useful but the create test data ckan command as far as I can see puts the publisher in the group table and while the package table has a nullable (i think) owner_org, so it would be handy to sense check from real data which is the right place to get the publisher/org id or name.

As long as we are happy to pick owner_id as the best bet we can implement that and see if we get any nulls in the dry run. Sound ok?

On the grouping of resources, no there isn't any but it's a csv so can be sorted by package or resource id outside this code. Likewise if we include org.

bin/python_scripts/script-runbook.md
```diff
@@ -0,0 +1,87 @@
+# Script runbook
+
+## Prerequisites
+
+- Set environment variable `POSTGRES_URL`
```
kentsanggds, line 5: Can probably add a note that when running the script on the container you can get the `POSTGRES_URL` from the `CKAN_SQLALCHEMY_URL`

bin/python_scripts/revert_link_deletions.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+
+import argparse
+import csv
+import logging
+import os
+import sys
+from datetime import UTC, datetime
+
+from python_scripts.check_links import Repository, setup_logging
```
kentsanggds, line 26: I can't run this in the `python_scripts` folder unless the `python_scripts` part is removed

bin/python_scripts/script-runbook.md
```diff
@@ -0,0 +1,87 @@
...
+| `WORKERS` | 50 | concurrent HTTP workers |
+| `HTTP_TIMEOUT` | `(5, 5)` | `(connect, read)` seconds |
+
+These are module constants near top of script. Change if needed according to tuning section above.
+
+### Outputs
+
+- `check_links.log` — run log (stable name, always current dir)
+- `check_links_report_{ts}.csv` — one row per resource (category + `to-delete` flag) - current dir unless `--output-dir` set
+- `packages_to_reindex_{ts}.txt` — unique package IDs that had a 404/410 — used as input to `solr_reindex_package_ids.py` - cuirrent dir unless `--output-dir` set
```
kentsanggds, line 60: The `solr_reindex_package_ids.py` was written when the CKAN framework was still running on pylons, since it was upgraded to use flask the paster commands will need to be updated to call `ckan search-index rebuild xxxx`

bin/python_scripts/script-runbook.md
```diff
@@ -0,0 +1,87 @@
+# Script runbook
```
kentsanggds, line 1: Can we also have a section to describe how to run the tests as they didn't seem to run for me.

bin/python_scripts/check_links.py
ashimali: Assumed owner_org although nullable will be the id to join group table on id.

Updated select on line 214

kentsanggds (changes requested):
I managed to get the script running after a few tweaks to the code which I have outlined in my comments. 

Also the script will have to be run on the pod, as port-forwarding from the pods is disabled probably due to security concerns.

bin/python_scripts/script-runbook.md
ashimali: does it take same input as solr_reindex_package_ids.py?

bin/python_scripts/revert_link_deletions.py
ashimali: you mean relative import?

bin/python_scripts/script-runbook.md
ashimali: And can you remove scripts that can't be used anymore from the repo.

ashimali: Actually why don't we use `cli.py:reindex_recent` instead as we update the package_metadata on 'delete' and 'revert'

bin/python_scripts/revert_link_deletions.py
kentsanggds: it wasn't working for me when I was running the command from within the container

bin/python_scripts/script-runbook.md
kentsanggds: it can take an input of ids, but the issue is that the process that it is calling needs to be updated

ashimali: Not sure what you mean. Can we pass the file to that command? Or do we need this at all?

So let's frame the question like this

1. What format would the input to ckan search-index rebuild xxxx be?
2. As an alternative, do we need it if we can use: cli.py:reindex_recent?

ashimali:
> I managed to get the script running after a few tweaks to the code which I have outlined in my comments.
> 
> Also the script will have to be run on the pod, as port-forwarding from the pods is disabled probably due to security concerns.

I assumed you'd be running it on the pod.

bin/python_scripts/script-runbook.md
kentsanggds: Basically this python file will need an update to run with the flask framework - https://github.com/alphagov/ckanext-datagovuk/blob/main/bin/python_scripts/solr_reindex_package_ids.py#L40

Alternatively you could just call a subprocess within your script to do a solr reindex?

kentsanggds (approved):
I've done a full test from check_links and revert and also have run the tests and it all appears to be working for me.

PR #2077, opened 2026-05-06
feat: add --verbose flag

ashimali:
Defaults to false

- If no flag passed then only deleted resources written to the csv report
- If flag is passed (no value after flag required) it sets to true and all resources checked written to csv report

Also added a note to script runbook about --limit flag which wasn't documented.


kentsanggds (approved):
LGTM

PR #2078, opened 2026-05-06
feat: sliding window

ashimali:
make script a bit more memory friendly by using sliding window over result rows and keeping pool of futures to process at around 200 futures in pool at any time.

Also scaled back retries count and interval. risk of rate limiting hosts slipping the net, but trade that off against getting an initial run done in reasonable time.

Ran local test with real data from earlier snapshot of production datasets and resources.

Using docker image with 4gb ram allocated ran in 4.5 hours with around 4% of urls rate limited by hosts, so won't have accurate results for those urls (approx 8k urls).

The 429s were clustered around 6 domains, so recommend to proceed with this and if required another following pr can address better handling of those hosts.

Also added a new column to csv output showing data.gov.uk page url for package/dataset.

### DIFF
--- a/bin/python_scripts/check_links.py
+++ b/bin/python_scripts/check_links.py
@@ -327,6 +327,7 @@ def run(
     reindex_path: str,
     mode: str,
     limit: int | None = None,
+    verbose: bool = False,
 ) -> None:
 
     # **NOTE**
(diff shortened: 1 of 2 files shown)


kentsanggds (approved):
LGTM

PR #2079, opened 2026-05-15
feat/fix: add created and last updated dates for resource.

ashimali:
Added created and last updated dates for resource row csv output.  Goal is to help triage of links that could be marked 'deleted'

Also fix bug where org name and id, not mapped through properly to csv! my fault.


kentsanggds (approved):
LGTM

PR #2082, opened 2026-05-29
Upload check links reports to s3

kentsanggds:
Add upload of reports to s3 after the check_links scripts completes and a download of reports on s3 so that they can be published.

This has been successfully tested on Integration and Staging running multiple CKAN pods.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-527

### DIFF
--- a/bin/python_scripts/lib/s3.py
+++ b/bin/python_scripts/lib/s3.py
@@ -28,5 +28,10 @@ class CkanOutputBucket(object):
         self.bucket.upload_file(path, s3_path)
 
 
-    def download_from_s3(self, s3_path):
-        self.bucket.download_file(s3_path)
+    def download_from_s3(self, s3_path, target_dir=None):
+        target_path = s3_path
+        if target_dir:
+            filename = s3_path.split("/")[-1]
+            target_path = target_dir + "/" + filename
+
+        self.bucket.download_file(s3_path, target_path)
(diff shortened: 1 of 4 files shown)


bin/python_scripts/lib/s3.py
```diff
@@ -0,0 +1,37 @@
...
+        bucket_name = os.environ.get('CKAN_OUTPUT_BUCKET_NAME')
+        if not bucket_name:
+            raise Exception("CKAN_OUTPUT_BUCKET_NAME environment variable is not set")
+
+        s3 = boto3.resource('s3')
+        self.bucket = s3.Bucket(bucket_name)
+
+    def get_s3_ls(self, path=None):
+        filenames = []
+        for obj in self.bucket.objects.all():
```
Obsiye, line 17: suggestion, feel free to ignore:

maybe not an issue now, `self.bucket.objects.all()` could end up being costly right? not sure how large the bucket will get but large buckets have caused slowness for me.

what if it looks in `check_links/TODAYS_DATE/ ` ? the new links exist there?

kentsanggds: The reports will be generated once a week, and might even be set to once a month later, this will be iterated on but for now I don't want to over-engineer it

Obsiye (approved):

PR #2086, opened 2026-06-10
feat: split dns and connection refused errors

ashimali:
This PR covers the following:

1. split dns and connection refused out from generic connection errors or by checking exception chain from requests. 
2. Increased number of retries 
3. Fixed bug on retry flag that swallowed final errors status codes for force retry list errors. This caused status code to not be recorded in csv for 5xx range errors, in verbose mode.


kentsanggds (approved):
LGTM

PR #2087, opened 2026-06-10
feat: apply link deletion script

ashimali:
feat: Single script for resource link deletions and reverting the deletions

Takes a report file and deletes, or reverts previously deleted flagged resources - writes list of resources to reindex as well as a copy of the input report named [report_name]_[deleted|activte].csv with additional columns 

- update-applied with value deleted|active
- update-applied-on with timestamp of delete or active updated

To apply to a subset of report, make a copy with just the rows to be deleted or reverted (i.e. marked active)


kentsanggds (approved):
Tried deletion and revertion on Integration

PR #2090, opened 2026-06-12
Add zscaler pem to dev.Dockerfile and set SSL env vars

kentsanggds:
This will allow the dev.Dockerfile container to make https requests

### DIFF
--- a/docker/ckan/dev.Dockerfile
+++ b/docker/ckan/dev.Dockerfile
@@ -1,4 +1,11 @@
-FROM --platform=$TARGETPLATFORM ghcr.io/alphagov/ckan:2.10
+FROM ghcr.io/alphagov/ckan:v1.18.2
+
+# copy zscaler cert and set env vars to use it for SSL connections
+# the zscaler cert should be available on your machine, otherwise follow this link to get it:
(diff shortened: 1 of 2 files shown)


docker/ckan/dev.Dockerfile
```diff
@@ -1,4 +1,9 @@
-FROM --platform=$TARGETPLATFORM ghcr.io/alphagov/ckan:2.10
+FROM ghcr.io/alphagov/ckan:v1.18.2
+
+# copy zscaler cert and set env vars to use it for SSL connections
+ADD zscaler.pem /usr/local/share/ca-certificates/zscaler.pem
```
currycoder, line 6: I notice that we aren't committing the certificate - which I suppose makes sense.  Is it worth something in the README to explain how to get it and where to put it?

kentsanggds: Sure, or we could commit it, is it safe to commit it? is it available publicly?

currycoder: great questions :D

kentsanggds: Decided to add some info about how to find the certs

currycoder (approved):

PR #2092, opened 2026-06-15
Adds a script to sort the check links report

kentsanggds:
Sorting by org name and http status code will make it easier to prepare the report for release to publishers

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-629

### DIFF
--- a/bin/python_scripts/check_links_sort.py
+++ b/bin/python_scripts/check_links_sort.py
@@ -1,0 +1,85 @@
+"""Sorts the check_links report CSV file by organization name and HTTP status."""
+
+import argparse
+import csv
+import logging
+import os
+import sys
+
+from lib.s3 import CkanOutputBucket
+from check_links import REPORT_HEADERS
+
+
+def setup_logging() -> logging.Logger:
+    logger = logging.getLogger(__name__)
+    if logger.handlers:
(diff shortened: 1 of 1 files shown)


bin/python_scripts/check_links_sort.py
```diff
@@ -0,0 +1,85 @@
+"""Sorts the check_links report CSV file by organization name and HTTP status."""
```
currycoder, line 1: Does this have to be put on to the pod/infrastructure?

I feel a bit like we are productionising something very early here.  And to run it we need to deploy it, jump on to a pod, exec the script, make it support S3 uploads.  It feels a bit nuts to sort a spreadsheet.

If it was just run locally on our machines for now, it would be quick to change and re-run.  And we could consider if we wanted to make it run on infrastructure later.

Obsiye (approved):

bin/python_scripts/check_links_sort.py
kentsanggds: The script still needs to be stored somewhere, and it will make the reports a bit easier to read, the script to run it after the check_links script is quite straightforward.

PR #2093, opened 2026-06-16
DGUK-621 Publisher emails associated with broken resource urls

Obsiye:
Create a python script that takes a csv with a list of `org-name` containing all the HTTP 404 and 410 error-ing resource links 

It then gets the publisher email relating to the org-name via the database and **outputs** a csv containing the emails and organisation names

ℹ️ By broken resource urls, I mean the url links in the dataset returning 404 or 410 http errors during the check links cronjob 

**TODO:**
- [X] Tested in integration
- [x] Add unit tests

### DIFF
--- a/bin/python_scripts/publisher_emails_of_broken_links.py
+++ b/bin/python_scripts/publisher_emails_of_broken_links.py
@@ -20,41 +20,39 @@ def setup_logging():
     logger.addHandler(c_handler)
 
 
-def read_resource_ids(csv_path):
-    ids = []
+def read_org_names(csv_path):
+    org_names = []
     with open(csv_path, 'r') as f:
         reader = csv.DictReader(f)
         for row in reader:
-            resource_id = row.get('resource-id', '').strip()
-            if resource_id:
-                ids.append(resource_id)
-    return ids
+            org_name = row.get('org-name', '').strip()
+            if org_name:
+                org_names.append(org_name)
+    return org_names
 
 
-def query_active_publisher_emails(connection, resource_ids):
+def query_active_publisher_emails(connection, group_names):
     cursor = connection.cursor()
-    placeholders = ', '.join(['%s'] * len(resource_ids))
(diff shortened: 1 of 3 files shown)


bin/python_scripts/publisher_emails_of_broken_links.py
```diff
@@ -0,0 +1,96 @@
...
+    SELECT DISTINCT u.email, g.name AS organisation
+    FROM "user" u
+    JOIN member m ON u.id = m.table_id AND m.table_name = 'user' AND m.state = 'active'
+    JOIN "group" g ON g.id = m.group_id
+    JOIN package p ON p.owner_org = g.id
+    JOIN resource r ON r.package_id = p.id AND r.state = 'active'
+    WHERE r.id IN ({ids})
+      AND u.state = 'active'
+    ORDER BY g.name, u.email
+    """.format(ids=placeholders)
```
currycoder: We had a chat in person about this, but suggest that this script consumes the organisation name from the spreadsheet and uses that to look up emails instead of going all the way through to the `resource.id`

Obsiye: thank you, i've updated the SQL. I'll test this in integration 🤞🏾

currycoder:
I have a general question for this one... How are we downloading the result CSV for this?  My understanding was that it was impossible to write to a local file on the pod and download it, which made all of the S3/persistent volumes things happen.  Might be a question for @kentsanggds

kentsanggds:
@currycoder the problem with the persistence on the pod is not that we couldn't download it, but that we couldn't persist it for other pods to use which were running on different nodes. In order to download this for processing the file can be copied from the pod using `kubectl cp ...` this file would then be fed into a Notify service to run against an email template to send to publishers.

kentsanggds (approved):
I'm going to approve this so that we have a script available if we need to generate another set of publisher contact emails for some major changes which will probably happen for other changes to the publisher

PR #2094, opened 2026-06-17
feat: add requests-ratelimiter adapter

ashimali:
Add requests-ratelimiter adapter as drop in for default httpadapter. set low max requests per sec per host.

We have a domains with high number of resources and therefore seeing a fair bit of rate limiting. using requests-ratelimiter we can drop in a replacment http adapter to play more nicely with those publishers by imposing host based rate limits. Limits set low for now, if runtime is dramatically slower we can test bumping it up in stepwise fashion by increasing max requests a second and/or upping the burst rate.

### DIFF
--- a/bin/python_scripts/check_links.py
+++ b/bin/python_scripts/check_links.py
@@ -26,8 +26,9 @@ from urllib.parse import urlsplit
 
 import psycopg2
 import requests
+
 from lib.s3 import CkanOutputBucket
-from requests.adapters import HTTPAdapter
+from requests_ratelimiter import LimiterAdapter
 from urllib3.util.retry import Retry
 
 LOG_FILE = "check_links.log"
@@ -38,6 +39,8 @@ HTTP_TIMEOUT = (5, 5)  # (connect, read) seconds
 WORKERS = 50
 MAX_INFLIGHT = WORKERS * 4
 HEAD_FALLBACK_STATUSES = {400, 403, 405, 501}
+PER_HOST_PER_SECOND = 1  # max reqs per/sec per host can increase if needed
+PER_HOST_BURST = 3  # up to 3 reqs per/sec burst allowed
 
 REPORT_HEADERS = [
     "datagovuk-url",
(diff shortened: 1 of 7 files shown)


kentsanggds (changes requested):
I don't think you have intended to add in the docker files as part of the PR, they should be removed

ashimali:
> I don't think you have intended to add in the docker files as part of the PR, they should be removed

Yeah I noticed that  and removed :)

bin/python_scripts/check_links.py
```diff
@@ -187,13 +189,18 @@ def session_factory(
...
+        status_forcelist=[429, 500, 502, 503, 504],
         allowed_methods=["HEAD", "GET"],
         backoff_factor=0.5,
         respect_retry_after_header=True,
         raise_on_status=False,
     )
-    adapter = HTTPAdapter(
+    # LimiterAdapter is an HTTPAdapter drop in, that maintains same retry
+    # config with in memory per host limit, no storage backend
+    adapter = LimiterAdapter(
```
ashimali, line 199: @kentsanggds  it seems in case of defra at least they are returning 403 for rate limits instead of 429. If you want to also add 403 to the rate limiter extension config, pass a keyword param `rate_limit_statuses=(403, 429)` to the constructor

kentsanggds: how will this also impact other orgs? if we treat all 403s as potential 429 does that mean that it will retry the 403 request after backing off, slowing down the report generation?

ashimali: I've added 403, so we can try and see what impact on runtime is.

kentsanggds: OK, I'll try this out on Integration with a limited run first

joelanman: does this mean only rate limiting once we get a 403 or 429? I don't think we should do that. We need to rate limit per server by default. We cannot rely on getting a 403 or 429. We relied on 429 and found that defra sent 403. Other servers may do other things or simply fall over and start 500ing. Also, by the time we get a 403 or 429, it could be too late and they just start blocking us. Defra have specifically told us to rate limit.

joelanman: sorry misread, my mistake, this means it goes even slower on a 403 or 439, not just on them

ashimali: No we already backoff and retry on a number of status codes, e..g. 429, 500, 502, 503, 504, however this is requests-ratelimiter saying in case of 403 and 429 there is additional *by host* limit to the base 2 requests per second at steady state although will allow burst of up to 5 request before throttling back to steady state.

ashimali: > sorry misread, my mistake, this means it goes even slower on a 403 or 439, not just on them

Messages crossed,

ashimali: worth noting that with the lib being used there are more levers that can be pulled. the current approach is an across the board rate limit policy (all hosts get same rate limiting config). It's possible to have a specific limit adapter for a specific host such that for example environmentagency.gov.uk could have a dedicated limit.

kentsanggds:
I've added a requirements.txt for check_links as it needs requests and request_ratelimiter

kentsanggds (approved):
LGTM

PR #2095, opened 2026-06-19
DGUK-641 Trim whitespace from resource url

Obsiye:
Trim the whitespace from the resource url before running the check links task

This prevents issues when calling the resource url to check if it is a broken link or  not


kentsanggds (approved):
LGTM

PR #2101, opened 2026-06-22
Upload the output report path to s3 so that it can be accessed by other pods

kentsanggds:
After processing a state change, upload the report to s3.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-499


amar-gds-ndl (approved):
LGTM

PR #2104, opened 2026-06-26
When reharvesting sources, report on the progress

kentsanggds:
Updated the harvest run with better logs to make it easier to see the progress.


Obsiye (approved):
LGTM

PR #2108, opened 2026-07-07
Prefix the moved scripts to old_ to avoid confusion

kentsanggds:
This should prevent some confusion for now whilst we decide how to handle the dependency in the process check links script


kentsanggds:
Not sure if we need to do this, check links work should be on the datagovuk-scripts repo

PR #2111, opened 2026-07-16
Raise error non 200 response

kentsanggds:
A publisher was experiencing issues with their harvest source but the error messages in the harvest jobs errors list did not make any sense and looked like the system was broken. 
There was in fact an issue with the publishers harvest source but the correct error didn't get propagated, so updating the Spatial sha should enable the surfacing of the correct error which was a 403 response to a server request on the harvest source. 

Additionally, the github actions was updated to show which build type is running as this wasn't clear in the github actions.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/727?selectedIssue=DGUK-763


amar-gds-ndl (approved):
LGTM

